PR #98, opened 2014-07-18
Use '.' instead of '=' in URLSafeBase64Encoder.

fletom:
According to [RFC 3986](http://www.ietf.org/rfc/rfc3986.txt) section 2.3, the equal sign (=) is not an
unreserved character in URLs:

> Characters that are allowed in a URI but do not have a reserved purpose are called unreserved.  These include uppercase and lowercase letters, decimal digits, hyphen, period, underscore, and tilde.

Since the tilde is common used to represent home directories, I chose to use the period (.) for padding instead.

### DIFF
--- a/src/nacl/encoding.py
+++ b/src/nacl/encoding.py
@@ -77,11 +77,11 @@ class URLSafeBase64Encoder(object):
 
     @staticmethod
     def encode(data):
-        return base64.urlsafe_b64encode(data)
+        return base64.urlsafe_b64encode(data).replace('=', '.')
 
     @staticmethod
     def decode(data):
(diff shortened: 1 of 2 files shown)


warner:
I think consistency with stdlib is more important here.. it's unfortunate that `base64` didn't replace the padding with a character that's actually reserved, but I expect that anyone who runs into that problem with pynacl's encoding function will have experienced it with stdlib first, and will be prepared to do whatever extra `.replace("=",".")` is necessary. Also, most implementations I've seen strip the padding anyways.

Closing as WONTFIX. Thanks!

PR #103, opened 2014-08-16
add type-checks to constructors

warner:
This also cleans up the imports in test_signing.py to match the convention used by the other test cases, and changes test_box.py to avoid the "pk"-vs-"sk" confusion by using "priv" and "pub".


warner:
I'm working on a replacement PR that's updated to current master, so I'm closing this one now.

PR #121, opened 2015-01-08
Upgrade libsodium to 1.0.1

reaperhulk:
Fixes #120 

As mentioned in that issue upgrading this just requires getting the latest tarball and replacing the libsodium directory. Arguably we shouldn't even untar it but since we currently do I have stayed consistent.

PR #124, opened 2015-03-04
disable flake8 on the two out-of-order "import nacl" lines

warner:
These imports must necessarily appear "out of order" (in violation of
flake8's rules), so disable flake8 checking for them without reducing
coverage of anything else.


warner:
(this was causing a travis failure on the "meta" builder. I suspect that flake8 has been updated since the last time the #121 branch was tested, which is why we didn't catch this earlier.

warner:
@alex this look good to you?

PR #126, opened 2015-03-04
restore docstrings for public, secret, and utils

warner:
We agreed in #99 that the docstrings were useful, although not intended
to be the primary documentation. This just copies them back from commit
ad97983 so the 0.3.0 release isn't so much of a regression for docstring
users.


reaperhulk:
Is this the last blocker for 0.3? I'll merge when tests are complete since this is just an agreed upon doc revert.

warner:
arguably, yeah. Although I have a batch of type-checking changes that I think might be worth of 0.3.0, I'll file a PR in a few minutes. (the changes were lounging in a PR for about 6 months, so I had to rebase and update everything).

Issue #43, opened 2013-10-25
`nacl.public.PrivateKey` class has no len() method


isislovecruft:
```
In [28]: box = nacl.secret.SecretBox(key)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-28-a37494c8a1fd> in <module>()
----> 1 box = nacl.secret.SecretBox(key)

/usr/local/lib/python2.7/dist-packages/PyNaCl-0.1.0-py2.7-linux-x86_64.egg/nacl/secret.pyc in __init__(self, key, encoder)
     47         key = encoder.decode(key)
     48
---> 49         if len(key) != self.KEY_SIZE:
     50             raise ValueError("The key must be exactly %s bytes long" %
     51                                 nacl.lib.crypto_secretbox_KEYBYTES)

TypeError: object of type 'PrivateKey' has no len()
```

Built from commit `fa71dfbea5496a9be8477c8c133845ed3eb05656` on branch `master`.

warner:
We've landed the type checks that ought to make errors like this give a useful message. I think we can close this now.

PR #127, opened 2015-03-04
add more type-checking to higher-level API calls

warner:
This should help with problems like #43 where passing the wrong arguments to an API call gives you a confusing error instead of a clear and immediate exception.

It also updates some tests to use arguments correctly, which is a useful teaching tool.


reaperhulk:
The checks seem reasonable to me but I'd like @dstufft to take a quick look.

dstufft:
+1 :shipit:

PR #128, opened 2015-03-04
Travis docker build experiment

reaperhulk:
This requires not actually installing libsodium in the normal path
when using "system". This may not work and may not be a good idea.


alex:
I think we can consider this expirement a failure :-)

PR #130, opened 2015-03-04
Add crypto pwhash scryptsalsa208sha256

malthe:


reaperhulk:
Looks like there's a py3 str/bytes issue to resolve in the tests.

alex:
I think this would be the first feature we're using that is from libsodium, vs. nacl itself?

malthe:
@reaperhulk doh ... fixed :-)

gleb-chipiga:
With with this pull request? Password hashing is very useful functionality

reaperhulk:
closing in favor of #204

PR #135, opened 2015-03-11
Enable checking keys for equality.

codethief:
I decided to use the `bytes` representation when comparing key objects. However, testing a key object and its own bytes representation for equality should not evaluate to `True`, see the corresponding unit tests.


codethief:
Oops, I didn't have my linter on (plus already deleted the original fork). Sorry, that'll teach me! I reposted the pull request as #136.

PR #136, opened 2015-03-12
Enable checking keys for equality.

codethief:
I decided to use the `bytes` representation when comparing key objects. However, testing a key object and its own bytes representation for equality should not evaluate to `True`, see the corresponding unit tests.


reaperhulk:
First, we owe you an apology for this ridiculously long PR time @codethief. This looks good, but would you mind implementing `__ne__` as well? Without that the `!=` comparison will always say False even when the keys are identical.

reaperhulk:
closed in favor of #208

PR #147, opened 2015-11-29
Add a getter for DH shared key

maqp:
I'm asking you to add a way for developers to also access the DH shared secret.

There might be other use cases too, but at least in [the one](https://github.com/maqp/tfc-cev) I'm working on where all key-exchange related data received needs to be manually typed on another computer, accessing the DH shared secret would make it much easier (and more secure) for the users when they don't also have to write additional nonce, ciphertext and tag to decrypt CSPRNG spawned shared secret key.

The reason I can't continuously use public key crypto box is lack of forward secrecy. So I want to use PyNaCl DH shared key, and refresh the shared secret key using HKDF after every message, creating a forward secret protocol similar to Silent Circle's SCIMP (and in some ways similar to Axolotl used in Signal).

Regarding code:
I'm sure you have your own variable naming policy (every class method seemed to be single word only, so feel free to change it and/or the functionality). I think it should also be verified that the shared key can be used on itself (and not e.g. after running it through hash / KDF) first.

-Markus

### DIFF
--- a/src/nacl/public.py
+++ b/src/nacl/public.py
@@ -82,29 +82,29 @@ class PrivateKey(encoding.Encodable, StringFixer, object):
         return self._private_key
 
     @classmethod
-    def generate(cls, add_e=''):
+    def generate(cls, ext_e=''):
         """
         Generates a random :class:`~nacl.public.PrivateKey` object
 
-        :param add_e: Additional entropy provided by user. XORed with system
-            entropy. If left empty, uses only nacl.utils.random().
+        :param ext_e: External entropy provided by user. XORed with system
+            entropy. If left empty, only nacl.utils.random() is used.
 
         :rtype: :class:`~nacl.public.PrivateKey`
         """
 
         size = nacl.bindings.crypto_box_SECRETKEYBYTES
 
-        if not add_e:
-            add_e = str(bytearray(size))
+        if not ext_e:
+            ext_e = str(bytearray(size))
 
-        if len(add_e) != size:
+        if len(ext_e) != size:
             raise ValueError(
-                "Additional entropy must be exactly %d bytes long" % size)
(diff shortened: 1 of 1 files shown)


reaperhulk:
An interesting use case. Seems reasonable, although we'll need documentation around this to explain what it is and why it's (usually) not a thing you need to touch.

It also looks like we've got failures due to unicode prefixing due to Python 3.2. Given that we dropped 3.2 support in cryptography and pyopenssl without issue I vote we nuke it here. @alex, @dstufft?

alex:
No objection to dropping 3.2, should be in a separate PR though.

On Sat, Nov 28, 2015 at 9:53 PM, Paul Kehrer <notifications@github.com>
wrote:

> An interesting use case. Seems reasonable, although we'll need
> documentation around this to explain what it is and why it's (usually) not
> a thing you need to touch.
>
> It also looks like we've got failures due to unicode prefixing due to
> Python 3.2. Given that we dropped 3.2 support in cryptography and pyopenssl
> without issue I vote we nuke it here. @alex <https://github.com/alex>,
> @dstufft <https://github.com/dstufft>?
>
> —
> <https://github.com/pyca/pynacl/pull/147#issuecomment-160362356>.
>



-- 

maqp:
@reaperhulk 
In the context of NaCl it's probably not something user needs to touch, but if someone wants to use a block cipher like AES together with ECDHE key exchange, they are going to need the getter. 

Also, I was able to write my [PoC implementation](https://imgur.com/gallery/KYfHR/) of forward secure secretbox with this modification.

reaperhulk:
@maqp would you mind writing some documentation with that information for this method?

maqp:
Making further changes to my master alters this pull request. So I'll make separate pull requests from new branches of new fork.

PR #148, opened 2015-11-29
remove py32 from our test matrix

reaperhulk:
we're not going to support it going forward


alex:
tox and setup.py as well

alex:
oh, needs an etry in the readme.rst changelog

PR #140, opened 2015-06-12
Switch to using CFFI 1.0 APIs

dstufft:

PR #154, opened 2015-12-06
Ignore all errors on mkdir in setup.py, because on FreeBSD it fails when directory already exists

yurivict:


setup.py
```diff
@@ -169,7 +169,7 @@ def run(self):
         # Ensure our temporary build directory exists
         try:
             os.makedirs(build_temp)
-        except IOError:
+        except:
```
alex: This doesn't look correct to me. What error is FreeBSD failing with, if not an `IOError`?

yurivict: OSError: [Errno 17] File exists: 'a'

alex: This code should catch `OSError` then, I'm not sure what conditions `IOError` would even be raised in.

yurivict: They are distinct, that's why I suggested IOError deletion.
IOError is when the genuine disk error occurs, or directory doesn't exist.

yurivict: Also: I made a port for FreeBSD. For some reason 'install' operation begins to build something. This normally doesn't happen with python ports. And python framework is very well behaved and stable in FreeBSD. So there is something strange with pynacl.

alex: If IOError really can occur, than this should catch the two of them. We are
not going to do a bare except though.

On Sun, Dec 6, 2015 at 9:58 AM, yurivict <notifications@github.com> wrote:

> In setup.py <https://github.com/pyca/pynacl/pull/154#discussion_r46771317>
> :
>
> > @@ -169,7 +169,7 @@ def run(self):
> >          # Ensure our temporary build directory exists
> >          try:
> >              os.makedirs(build_temp)
> > -        except IOError:
> > +        except:
>
> Also: I made a port for FreeBSD. For some reason 'install' operation
> begins to build something. This normally doesn't happen with python ports.
> And python framework is very well behaved and stable in FreeBSD. So there
> is something strange with pynacl.
>
> —
> <https://github.com/pyca/pynacl/pull/154/files#r46771317>.
>



-- 

yurivict: You shouldn't catch anything there. Any error is the real error.
Just test the directory exists instead. The other errors are all fatal.

dstufft: Testing the directory exists has a race condition. What happens if the directory starts to exist between when you did the check if it exists and when you attempt to create it.

#140 should fix the thing where ``install`` builds stuff.

alex: We absolutely odn't want to catch random errors, if there's a `NameError` for example, that indicates a bug, and we don't want to silence this.

yurivict: Agree. Then you need to only ignore the errno=EEXIST

Renelvon: According to the [latest docs](https://docs.python.org/3.5/library/os.html#os.makedirs), an `OSError` is raised if the directory exists, so the code should probably read
```
except OSError:
```
Moreover, `IOError` doesn't exist since Python 3.3; it has been [merged](https://docs.python.org/3.4/library/exceptions.html#OSError) into `OSError`. This is not the case in Python 2.x, where `IOError` and `OSError` are separate exceptions. However, this shouldn't matter, because [`os.makedirs`](https://docs.python.org/2.6/library/os.html#os.makedirs) also raises `OSError` in Python 2.6 and 2.7.

It appears to me that catching an `IOError` is a mistake and catching `OSError` is the correct response, given the code that rests inside the `try` block. However, if there is some peculiar reason for catching `IOError` anyway, one can do so easily:
```
except (OSError, IOError):
```
This should work in Python 2.6+ and Python 3.3+, which is enough for this library's purpose.

PR #155, opened 2015-12-12
Switch to CFFI 1.0 APIs (Redux)

reaperhulk:
Taking over #140

PR #156, opened 2015-12-12
Travis config

reaperhulk:
pull out travis config changes from #155


.travis/install.sh
```diff
@@ -0,0 +1,26 @@
...
+    cd libsodium-*
+    ./configure
+    make
+    make check
+    sudo make install
+    sudo ldconfig
+fi
+
+if [[ "${TOXENV}" == "pypy" ]]; then
+    git clone https://github.com/yyuu/pyenv.git ~/.pyenv
```
alex: I think you might need to replace this with a `git pull` in `~/.pyenv`, the dir already exists.

PR #157, opened 2015-12-22
Signing.py: Name and hide {Public,Private}Key import to avoid confusion

JackWink:
From @warner's feedback, we'll hide these imports to avoid accidental use.

### DIFF
--- a/src/nacl/signing.py
+++ b/src/nacl/signing.py
@@ -102,7 +102,7 @@ class VerifyKey(encoding.Encodable, StringFixer, object):
 
         return nacl.bindings.crypto_sign_open(smessage, self._key)
 
-    def to_public_key(self):
+    def to_curve25519_public_key(self):
         """
         Converts a :class:`~nacl.signing.VerifyKey` to a
         :class:`~nacl.public.PublicKey`
@@ -183,7 +183,7 @@ class SigningKey(encoding.Encodable, StringFixer, object):
 
         return SignedMessage._from_parts(signature, message, signed)
 
-    def to_private_key(self):
+    def to_curve25519_private_key(self):
         """
         Converts a :class:`~nacl.signing.SigningKey` to a
         :class:`~nacl.public.PrivateKey`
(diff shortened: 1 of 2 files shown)


warner:
Looks good, thanks for jumping on this one!

warner:
.. although, now that I think about it, the method names raise a similar concern:

```python
signer = SigningKey.generate()
x = signer.to_private_key() # what is x? "private signing key"?
```

Could we maybe rename `to_private_key()` to `to_curve25519_private_key()`, etc?

Sorry to be so worried about this.. I 've tried personally to get into the habit of alyways saying "signing key" and "verifying key" when referring to the private/public parts of signature systems, so I guess having some sort of crossover between the two naming regimes makes me pay extra-close attention to potential confusion.

JackWink:
I think it makes sense to do all we can to prevent user error. I was initially in favor of something like `to_encryption_private_key()` over `to_curve25519_private_key()` since this would be the first place the public API specifically states the implementation details, but the docs are pretty heavy with ed25519/curve25519 references so I'm not sure that would be a real concern. I'd be curious to get the project maintainers' opinion, but I'm fine to go with the `to_curve25519_` prefix for now.

warner:
Yeah, I've got mixed feelings about hiding the implementation name. On one hand:

* Short names are much nicer to work with
* Downstream application code would look pretty ugly if every `encrypt()` call had to mention `salsa20poly1305curve25519etcetcetc` too
* Up-up-stream code (djb's nacl library) provides short aliases

On the other hand:

* If you actually used the nacl aliases, then if nacl's default algorithm choices were ever changed, your program would suddenly lose compatibility with older versions. I'm into encrypted storage applications with long-term reliability, so this is a big deal.
* Trying to share keys between two very different operations is a pretty advanced feature, and only possible if the two algorithms are deeply related to one another. You couldn't do it with, say, RSA encryption keys and DSA verifying keys. So downstream application authors who are doing that should really know more details than usual.

How does our (up*1)-stream provider (libsodium) name these functions? I don't know if they've gone through the same thought process as us.

JackWink:
Sorry I dropped off for a while, just noticed your feedback and that this got added to v0.4.0. @warner that makes sense, we'll go with `to_curve25519_{public, private}_key()`.

warner:
Looks good to me, landing..

PR #159, opened 2016-01-01
Add getter for Curve25519 shared secret

maqp:
Moving the discussion from #147 to here. 

The pull request is now under it's own branch. @reaperhulk asked me to write some documentation so I added a section to public.rst plus a docstring, that warn user to use a symmetric cipher with nonce and 32-byte keys.

### DIFF
--- a/docs/public.rst
+++ b/docs/public.rst
@@ -167,3 +167,15 @@ Reference
         :param encoder: A class that is able to decode the plaintext.
 
         :return bytes: The decrypted plaintext.
+
+    .. method:: shared_key()
+
+        Returns the Curve25519 shared secret, that can then be used as a key in
+        other symmetric ciphers.
+
+        .. warning:: It is **VITALLY** important that you use a nonce with your
+            symmetric cipher. If you fail to do this, you compromise the
+            privacy of the messages encrypted. Ensure that the key length of
+            your cipher is 32 bytes.
+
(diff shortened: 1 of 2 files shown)


lmctv:
@maqp I think you are more qualified than myself to think about the missing unit test(s) for the shared_key() method.

maqp:
I'm unsure how many are needed. Accessing the attribute wouldn't have required a getter in Python but it's nice to have the usability of the attribute documented and having the getter makes the API more beautiful.

But if they are really desired, what could be added is type and length check, plus a test vector where known private and public key produce a known shared secret:

https://tools.ietf.org/id/draft-josefsson-tls-curve25519-02.html has following test vectors

    Sk_a = 5AC99F33632E5A768DE7E81BF854C27C46E3FBF2ABBACD29EC4AFF517369C660
    Pk_a = 057E23EA9F1CBE8A27168F6E696A791DE61DD3AF7ACD4EEACC6E7BA514FDA863
    Sk_b = 47DC3D214174820E1154B49BC6CDB2ABD45EE95817055D255AA35831B70D3260
    Pk_b = 6EB89DA91989AE37C7EAC7618D9E5C4951DBA1D73C285AE1CD26A855020EEF04
    SS   = 61450CD98E36016B58776A897A9F0AEF738B99F09468B8D6B8511184D53494AB

PR #160, opened 2016-01-01
Add external entropy source for private key generation

maqp:
This commit adds an optional parameter for user to easily mix in entropy from external sources:

    ext_entropy = usersOwnEntropyGenerator()
    skbob = PrivateKey.generate(ext_entropy)

If additional entropy is not provided--

    skbob = PrivateKey.generate()

the external key is replaced with zero-bit string that has equal length to PrivateKey.

The mixing itself is done with a simple constant time XOR operation. External entropy is defined before library's own random bits are generated with nacl.utils.random(), thus no weak by accident (or malice) entropy source can negatively affect the quality of final key.

Checks for input length and type ensure user can't misuse the function. I also added a docstring for the variable and documentation for readthedocs.

### DIFF
--- a/docs/public.rst
+++ b/docs/public.rst
@@ -112,9 +112,12 @@ Reference
         An instance of :class:`~.nacl.public.PublicKey` that corresponds with
         the private key.
 
-    .. classmethod:: generate()
+    .. classmethod:: generate(ext_e)
 
         Generates a random :class:`~nacl.public.PrivateKey` object
+
+        :param ext_e: Optional external entropy provided by user. XORed with
+            system entropy. If empty, only nacl.utils.random() is used.        
 
         :return: An instance of :class:`~nacl.public.PrivateKey`.
 
--- a/src/nacl/public.py
+++ b/src/nacl/public.py
@@ -82,14 +82,35 @@ class PrivateKey(encoding.Encodable, StringFixer, object):
         return self._private_key
 
     @classmethod
-    def generate(cls):
+    def generate(cls, ext_e=''):
         """
         Generates a random :class:`~nacl.public.PrivateKey` object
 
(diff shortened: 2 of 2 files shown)


maqp:
It would appear I can't fix the pull request code to get rid of errors. Changes need to be made into test_box.py itself. I don't feel confident making such edits so I could really use some help.

alex:
What's the use case for this? IMO this is outside the spirit of nacl, where things are supposed to have simple APIs with no extraneous parameters.

maqp:
The program I'm implementing PyNaCl for can load entropy from free hardware design HWRNG, connected to GPIO of SoC such as Raspberry Pi. When the software is run on other hardware platform, the HWRNG can still be queried from RasPi with Paramiko/SSH. This is to ensure security especially with live systems where urandom's entropy pool can be low after boot.

But I understand your point of view. In most applications urandom works just fine. I'll see if I can figure out a way to feed the HWRNG entropy directly to kernel, that way no changes need to be made.

alex:
Yes, feeding the `HWRNG` directly into the kernel is definitely a better solution.

maqp:
I'll close this for now. If you have time @alex , could you review the shared secret getter at #159?

maqp:
To anyone who might stumble on this issue in the future, additional entropy can be mixed with e.g.

    import nacl.utils
    import nacl.public

    def get_external_entropy():
        return bytes(32)  # This function should return the 32 bytes of external entropy

    def xor(string1, string2):
        assert len(string1) == len(string2)
        return b''.join([bytes([b1 ^ b2]) for b1, b2 in zip(string1, string2)])
        

    sk_bob = nacl.public.PrivateKey(xor(get_external_entropy(),
                                        nacl.utils.random(nacl.public.PrivateKey.SIZE)))

PR #164, opened 2016-01-09
preliminary windows support via some new flags

reaperhulk:
PYNACL_SODIUM_LIBRARY_NAME lets you set what it should be trying to link against.

PYNACL_SODIUM_STATIC will define SODIUM_STATIC before include to set the proper dllexport visibility for static linking on windows.

This PR also should be tested by a new jenkins job... let's see if it works.

### DIFF
--- a/src/bindings/build.py
+++ b/src/bindings/build.py
@@ -14,7 +14,7 @@
 from __future__ import absolute_import, division, print_function
 
 import glob
-import os.path
+import os
 
 from cffi import FFI
 
@@ -34,6 +34,17 @@ for header in HEADERS:
     with open(header, "r") as hfile:
         ffi.cdef(hfile.read())
 
+# we need to easily control a few values for Windows builds
+source = []
+if os.getenv("PYNACL_SODIUM_STATIC") is not None:
+    source.append("#define SODIUM_STATIC")
+
+source.append("#include <sodium.h>")
+
+if os.getenv("PYNACL_SODIUM_LIBRARY_NAME") is not None:
+    libraries = [os.getenv("PYNACL_SODIUM_LIBRARY_NAME")]
+else:
+    libraries = ["sodium"]
 
 # Set our source so that we can actually build our bindings to sodium.
-ffi.set_source("_sodium", "#include <sodium.h>", libraries=["sodium"])
+ffi.set_source("_sodium", "\n".join(source), libraries=libraries)
--- a/src/bindings/build.py
+++ b/src/bindings/build.py
@@ -14,7 +14,8 @@
 from __future__ import absolute_import, division, print_function
 
 import glob
-import os
+import os.path
+import sys
 
 from cffi import FFI
 
@@ -41,8 +42,8 @@ if os.getenv("PYNACL_SODIUM_STATIC") is not None:
 
 source.append("#include <sodium.h>")
 
-if os.getenv("PYNACL_SODIUM_LIBRARY_NAME") is not None:
-    libraries = [os.getenv("PYNACL_SODIUM_LIBRARY_NAME")]
+if sys.platform == "windows":
+    libraries = ["libsodium"]
 else:
     libraries = ["sodium"]
 
--- a/src/bindings/build.py
+++ b/src/bindings/build.py
@@ -35,14 +35,18 @@ for header in HEADERS:
     with open(header, "r") as hfile:
         ffi.cdef(hfile.read())
 
-# we need to easily control a few values for Windows builds
 source = []
+
+# SODIUM_STATIC controls the visibility of symbols in the headers. (see
+# export.h in the libsodium source tree). If you do not set SODIUM_STATIC
+# when linking against the static library in Windows then the compile will
+# fail with no symbols found.
 if os.getenv("PYNACL_SODIUM_STATIC") is not None:
     source.append("#define SODIUM_STATIC")
 
 source.append("#include <sodium.h>")
 
-if sys.platform == "windows":
+if sys.platform == "win32":
     libraries = ["libsodium"]
 else:
     libraries = ["sodium"]
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,7 @@ deps =
     pytest
     coverage
     pretend
-passenv = PYNACL_SODIUM_STATIC LIB INCLUDE
+passenv = SODIUM_INSTALL PYNACL_SODIUM_STATIC LIB INCLUDE
 commands =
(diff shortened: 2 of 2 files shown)


src/bindings/build.py
```diff
@@ -34,6 +34,17 @@
...
 
+# we need to easily control a few values for Windows builds
+source = []
+if os.getenv("PYNACL_SODIUM_STATIC") is not None:
+    source.append("#define SODIUM_STATIC")
+
+source.append("#include <sodium.h>")
+
+if os.getenv("PYNACL_SODIUM_LIBRARY_NAME") is not None:
+    libraries = [os.getenv("PYNACL_SODIUM_LIBRARY_NAME")]
```
alex: Why would someone need a custom name?

reaperhulk:
jenkins, retest this please

src/bindings/build.py
```diff
@@ -34,6 +34,17 @@
     with open(header, "r") as hfile:
         ffi.cdef(hfile.read())
 
+# we need to easily control a few values for Windows builds
+source = []
+if os.getenv("PYNACL_SODIUM_STATIC") is not None:
+    source.append("#define SODIUM_STATIC")
```
alex: Can you add a comment explaining that what this does is cause `libsodium` to be a static lib... or whatever this define does?

src/bindings/build.py
reaperhulk: The Windows library naming scheme makes this somewhat useful. When you ask for "sodium" on linux/OS X it gets translated to "libsodium.so" or "libsodium.dylib" but on windows it just looks for "sodium.lib". The libsodium precompiled libraries, however, call it "libsodium" on windows. We could just rename it to "sodium" but it's nice to be able to use the libsodium.org download without renaming a bunch of files.

Alternately we could just define the library name to be "libsodium" on Windows, but that does mean windows users (at least the ones who want to compile it themselves) need to know that we require the library to be named that.

I'm neutral on which approach we take here (although there is a reasonable argument to be made that it's unlikely anyone on Windows is going to name it anything other than "libsodium")

alex: Sorry, if the libsodium binaries name it `libsodium`, why can't we just write `"libsodium"` here and be happy?

reaperhulk: If you compile it yourself it won't necessarily be named that, but as I said before, if anyone is compiling it themselves it is perhaps reasonable to state that they need to learn what the expected name is for linking.

alex: Why wouldn't it be named that -- is there a different default for
self-compiled, no default, or just "someone might change it"?

On Sat, Jan 9, 2016 at 10:53 AM, Paul Kehrer <notifications@github.com>
wrote:

> In src/bindings/build.py
> <https://github.com/pyca/pynacl/pull/164#discussion_r49265121>:
>
> > @@ -34,6 +34,17 @@
> >      with open(header, "r") as hfile:
> >          ffi.cdef(hfile.read())
> >
> > +# we need to easily control a few values for Windows builds
> > +source = []
> > +if os.getenv("PYNACL_SODIUM_STATIC") is not None:
> > +    source.append("#define SODIUM_STATIC")
> > +
> > +source.append("#include <sodium.h>")
> > +
> > +if os.getenv("PYNACL_SODIUM_LIBRARY_NAME") is not None:
> > +    libraries = [os.getenv("PYNACL_SODIUM_LIBRARY_NAME")]
>
> If you compile it yourself it won't necessarily be named that, but as I
> said before, if anyone is compiling it themselves it is perhaps reasonable
> to state that they need to learn what the expected name is for linking.
>
> —
> <https://github.com/pyca/pynacl/pull/164/files#r49265121>.
>



-- 

reaperhulk: Since there's no suffix difference for static vs dynamic vs link time compile optimization library types Windows people appear to sometimes totally rename their libraries. We saw this with the slproweb OpenSSL library names before we switched to our own for cryptography. In this case there are vcxproj files (at least in the git tree, they don't appear to actually ship in the tarball) for the various VS201X releases that probably do give the "libsodium" name by default.

alex: Ok, I think just hardcode it to `libsodium` and we can do something
different if someone complains.

On Sat, Jan 9, 2016 at 10:59 AM, Paul Kehrer <notifications@github.com>
wrote:

> In src/bindings/build.py
> <https://github.com/pyca/pynacl/pull/164#discussion_r49265169>:
>
> > @@ -34,6 +34,17 @@
> >      with open(header, "r") as hfile:
> >          ffi.cdef(hfile.read())
> >
> > +# we need to easily control a few values for Windows builds
> > +source = []
> > +if os.getenv("PYNACL_SODIUM_STATIC") is not None:
> > +    source.append("#define SODIUM_STATIC")
> > +
> > +source.append("#include <sodium.h>")
> > +
> > +if os.getenv("PYNACL_SODIUM_LIBRARY_NAME") is not None:
> > +    libraries = [os.getenv("PYNACL_SODIUM_LIBRARY_NAME")]
>
> Since there's no suffix difference for static vs dynamic vs link time
> compile optimization library types Windows people appear to sometimes
> totally rename their libraries. We saw this with the slproweb OpenSSL
> library names before we switched to our own for cryptography. In this case
> there are vcxproj files (at least in the git tree, they don't appear to
> actually ship in the tarball) for the various VS201X releases that probably
> do give the "libsodium" name by default.
>
> —
> <https://github.com/pyca/pynacl/pull/164/files#r49265169>.
>



-- 

src/bindings/build.py
```diff
@@ -34,6 +35,17 @@
...
+
+source.append("#include <sodium.h>")
+
+if sys.platform == "windows":
```
alex: pretty sure this should be `"win32"`

reaperhulk: yup

src/bindings/build.py
```diff
@@ -34,6 +35,17 @@
     with open(header, "r") as hfile:
         ffi.cdef(hfile.read())
 
+# we need to easily control a few values for Windows builds
```
alex: This comment no longe really makes sense, I think you can remove it and replace with just explaining the sodium static bit.

reaperhulk:
jenkins retest this please

reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, retest this please

PR #170, opened 2016-01-23
add aes256gcm support

xueyumusic:
AES256 GCM mode is a popular mode of operation for the symmetric key cryptographic block ciphers. Support aes256 gcm mode could be beneficial to many applications.

### DIFF
--- a/src/bindings/crypto_aead_aes256gcm.h
+++ b/src/bindings/crypto_aead_aes256gcm.h
@@ -1,0 +1,27 @@
+
+size_t crypto_aead_aes256gcm_keybytes();
+size_t crypto_aead_aes256gcm_nsecbytes();
+size_t crypto_aead_aes256gcm_npubbytes();
+size_t crypto_aead_aes256gcm_abytes();
+
+
+int crypto_aead_aes256gcm_encrypt(unsigned char *c,
+                                  unsigned long long *clen_p,
+                                  const unsigned char *m,
+                                  unsigned long long mlen,
+                                  const unsigned char *ad,
+                                  unsigned long long adlen,
+                                  const unsigned char *nsec,
+                                  const unsigned char *npub,
+                                  const unsigned char *k);
+
+
+int crypto_aead_aes256gcm_decrypt(unsigned char *m,
+                                  unsigned long long *mlen_p,
+                                  unsigned char *nsec,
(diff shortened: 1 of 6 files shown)


reaperhulk:
I'm a bit conflicted here. On the one hand AES GCM is clearly a valuable AEAD due to its wide support -- on the other libsodium is generally meant to be used for djb primitives (and as such chacha20poly1305 is probably a better AEAD choice). The AES GCM implementation in libsodium also is only supported on processors that support hardware acceleration of AES (AES-NI) and carryless multiplication (PCLMULQDQ), so it doesn't have any side channel issues, but also has significant compatibility concerns.

@warner do you have an opinion on this? Exposing this would definitely take us more towards the libsodium side rather than the nacl one.

alex:
I don't understand why libsodium exposes AES256GCM at all. To the maximum
extent permitted by law, pynacl should be about not giving users choices
IMO.

On Mon, Jan 25, 2016 at 10:28 AM, Paul Kehrer <notifications@github.com>
wrote:

> I'm a bit conflicted here. On the one hand AES GCM is clearly a valuable
> AEAD due to its wide support -- on the other libsodium is generally meant
> to be used for djb primitives (and as such chacha20poly1305 is probably a
> better AEAD choice). The AES GCM implementation in libsodium also is only
> supported on processors that support hardware acceleration of AES (AES-NI)
> and carryless multiplication (PCLMULQDQ), so it doesn't have any side
> channel issues, but also has significant compatibility concerns.
>
> @warner <https://github.com/warner> do you have an opinion on this?
> Exposing this would definitely take us more towards the libsodium side
> rather than the nacl one.
>
> —
> <https://github.com/pyca/pynacl/pull/170#issuecomment-174543518>.
>



-- 

dstufft:
I agree with @alex I believe. There are plenty of libraries out there that will give you a grab bag of primitives to do whatever you want with, adding another one doesn't seem useful.

tiran:
@dstufft @alex :+1: 

This feature request is Pandora's box. Don't open the box.

dstufft:
d654bc1d15a0295553a31ef60e51bbde7e6c753a4c3acac2793cd2e0045aa3cf

Lukasa:
:+1: to @alex. Choices are a bad idea for pynacl.

PR #171, opened 2016-01-24
fix absolute paths in the SOURCES.txt when generating an sdist

reaperhulk:
These absolute paths didn't stop installation from sdist, but prevented making a wheel.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,12 +51,20 @@ else:
 
 
 def here(*paths):
+    return os.path.relpath(
+        abshere(*paths),
+        os.path.abspath(os.path.dirname(__file__)),
+    )
+
+
+def abshere(*paths):
     return os.path.abspath(os.path.join(os.path.dirname(__file__), *paths))
+
 
 sodium = functools.partial(here, "src/libsodium/src/libsodium")
 
 
-sys.path.insert(0, here("src"))
+sys.path.insert(0, abshere("src"))
 
 
 import nacl # flake8: noqa
@@ -169,7 +177,7 @@ class build_clib(_build_clib):
             os.chmod(here(filename), 0o755)
 
         # Locate our configure script
-        configure = here("src/libsodium/configure")
+        configure = abshere("src/libsodium/configure")
 
         # Run ./configure
         subprocess.check_call(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -51,12 +51,20 @@
...
 
 def here(*paths):
+    return os.path.relpath(
+        abshere(*paths),
+        os.path.abspath(os.path.dirname(__file__)),
+    )
+
+
+def abshere(*paths):
     return os.path.abspath(os.path.join(os.path.dirname(__file__), *paths))
```
alex: Err, wouldn't it make more sense to have `abshere` be implemented in terms of `here`?

setup.py
```diff
@@ -51,12 +51,17 @@
 
 
 def here(*paths):
-    return os.path.abspath(os.path.join(os.path.dirname(__file__), *paths))
+    return os.path.relpath(os.path.join(*paths))
```
alex: Is the `relpath` needed?

reaperhulk: If we want here to join any path (even if it's already an absolute) and provide a path relative to the cwd then yes. Otherwise we should just remove here entirely? (Removing here entirely probably makes the most sense)

alex: `os.path.dirname(__file__)` will always be relative in this context, right?

On Sat, Jan 23, 2016 at 9:22 PM, Paul Kehrer <notifications@github.com>
wrote:

> In setup.py <https://github.com/pyca/pynacl/pull/171#discussion_r50629213>
> :
>
> > @@ -51,12 +51,17 @@
> >
> >
> >  def here(*paths):
> > -    return os.path.abspath(os.path.join(os.path.dirname(__file__), *paths))
> > +    return os.path.relpath(os.path.join(*paths))
>
> If we want here to join any path (even if it's already an absolute) and
> provide a path relative to the cwd then yes. Otherwise we should just
> remove here entirely? (Removing here entirely probably makes the most sense)
>
> —
> <https://github.com/pyca/pynacl/pull/171/files#r50629213>.
>



-- 

dstufft: I think ``__file__`` is an absolute path.

PR #183, opened 2016-04-25
Fixed internal rST reference in secret documentation

jezdez:


alex:
Nice catch! Will merge as soon as CI finishes.

docs/secret.rst
```diff
@@ -81,7 +81,7 @@ reason, many protocols derive a new key for each session, reset the counter
 to zero with each new key, and never store the derived key or the counter.
 
 You can safely generate random nonces by calling
-:class:``~nacl.utils.random(SecretBox.NONCE_SIZE)``.
+:class:`~nacl.utils.random(SecretBox.NONCE_SIZE)`.
```
alex: I'm not sure this is going to work actually, the construction is nonsensical, it's using `:class:`, but the thing it's showing is a function call. I think you actaulyl want to remove the `:class:` and leave the double backticks.

jezdez: That's a fair point, d'oh! Since `nacl.utils.random` is used a few places in the documentation, should this be a documented function in `docs/utils.rst` that we can link to from here?

alex: That's probably a good idea, but if you don't feel like doing all that
work, just the backticks so it renders correctly is good enough for me

On Mon, Apr 25, 2016 at 12:36 PM, Jannis Leidel <notifications@github.com>
wrote:

> In docs/secret.rst
> <https://github.com/pyca/pynacl/pull/183#discussion_r60944984>:
>
> > @@ -81,7 +81,7 @@ reason, many protocols derive a new key for each session, reset the counter
> >  to zero with each new key, and never store the derived key or the counter.
> >
> >  You can safely generate random nonces by calling
> > -:class:``~nacl.utils.random(SecretBox.NONCE_SIZE)``.
> > +:class:`~nacl.utils.random(SecretBox.NONCE_SIZE)`.
>
> That's a fair point, d'oh! Since nacl.utils.random is used a few places
> in the documentation, should this be a documented function in
> docs/utils.rst that we can link to from here?
>
> —
> You are receiving this because you commented.
> <https://github.com/pyca/pynacl/pull/183/files/c65a094f72bc119480025a8ecafb00becc454cc8#r60944984>
>



-- 
GPG Key fingerprint: D1B3 ADC0 E023 8CA6

jezdez: Nah, no work at all, pushed an update..

reaperhulk:
👍

Issue #186, opened 2016-05-14
sodium_init() fails because already initialized


asn-the-goblin-slayer:
Hello,

I'm using PyNaCl in a new secure messaging project: https://github.com/asn-the-goblin-slayer/viola

I'm experiencing problems with `sodium_core.py:sodium_init()` failing to initialize sodium, because apparently it has already been initialized. Specifically it hits the segment below:

```
def sodium_init():
    """
    Initializes sodium, picking the best implementations available for this
    machine.
    """
    retval = lib.sodium_init()
    if retval != 0:
        raise CryptoError("Could not initialize sodium %d" % retval)
```

and `lib.sodium_init()` returns `1` because it has already been initialized somehow. I think this might have to do
with the way I import nacl, and the way pynacl initializes libsodium without checking if it has already been inited.

Unfortunately, while I could reproduce this issue in the past, I can't anymore. But I just received a bug report about it:
https://github.com/asn-the-goblin-slayer/viola/issues/2

Wouldn't it be more appropriate to not call `sodium_init()` if it has already been initialized? Or alternatively, to not raise an exception if `lib.sodium_init()` returns 1 (already initialized)?

Or are there any guidelines on how to import nacl modules so that this doesn't happen? 
For example here is how I import nacl submodules in my project:

```
./viola/crypto.py:8:import nacl.signing
./viola/crypto.py:9:import nacl.encoding
./viola/crypto.py:10:import nacl.secret
./viola/crypto.py:11:import nacl.utils
./viola/viola.py:16:import nacl.exceptions # XXX dirty
./viola/room.py:3:import nacl.utils # XXX dirty. only crypto should touch nacl. move there!!!
./viola/introduction.py:10:import nacl.exceptions # XXX dirty
```

This issue has been quite annoying while deploying viola to other systems :(
It would be great if it could be resolved somehow!

Thanks!

asn-the-goblin-slayer:
Hello, I also provided a patch here: 
https://github.com/pyca/pynacl/pull/187

lmctv:
@asn-the-goblin-slayer I think this could be closed, if #199 works for you too.

asn-the-goblin-slayer:
Yep, closing this. Thanks!

PR #187, opened 2016-05-14
Don't fail if sodium is already initialized.

asn-the-goblin-slayer:
No reason to fail if sodium is already initialized.

From the sodium docs (https://download.libsodium.org/doc/usage/index.html):

```
sodium_init() initializes the library and should be called before any other function provided by Sodium. The function can be called more than once, but it should not be executed by multiple threads simultaneously. Add appropriate locks around the function call if this scenario can happen in your application.
```

This should address https://github.com/pyca/pynacl/issues/186 .

### DIFF
--- a/src/nacl/bindings/sodium_core.py
+++ b/src/nacl/bindings/sodium_core.py
@@ -22,5 +22,5 @@ def sodium_init():
     Initializes sodium, picking the best implementations available for this
     machine.
     """
-    if lib.sodium_init() != 0:
+    if lib.sodium_init() == -1:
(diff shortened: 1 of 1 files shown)


JackWink:
Should this also add locking around the init function?

asn-the-goblin-slayer:
Hey Jack, not sure if it's worth adding locking there. Depends on how PyNaCl handles threads.

Also, it seems that my patch broke the travis checks? :-1: :(

reaperhulk:
Resolved with #199

PR #195, opened 2016-08-16
Set default python interpreter version to 3.5

lmctv:
This avoid tox errors on TOXENV=py35 on newly-created (container-based) travis builds


reaperhulk:
It doesn't look like this is necessary with the non-container builds for now?

Issue #194, opened 2016-08-16
Travis tests fail with "sudo: must be setuid root"


lmctv:
Travis check log files end with

   sudo make install
   sudo: must be setuid root
   The command ".travis/install.sh" failed and exited with 1 during .

Since container-based execution environment became the default on trevis-ci

lmctv:
Closing since the merge of #197 resolves the issue

PR #196, opened 2016-08-16
Remove SODIUM_INSTALL=system lines

lmctv:
To overcome the need for sudo in the build environment. See:  #194


lmctv:
This is not needed if #197 gets merged. The main gain from this pull request is getting quicker completing travis tests.

reaperhulk:
We need to revisit the system vs bundled approach, but for now I've merged the `sudo: required` patch to unblock the CI.

lmctv:
@reaperhulk  you're right, the main interpreter version bump is needed only when running on container-based infrastructure. Let me know if you prefer to keep both this and #195 open as a reminder or close them. Thank you

PR #198, opened 2016-08-16
Update pypy

lmctv:
Instead of staying on pypy (and cffi) versions from october/november 2015

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -20,8 +20,8 @@ if [[ "${TOXENV}" == "pypy" ]]; then
     PYENV_ROOT="$HOME/.pyenv"
     PATH="$PYENV_ROOT/bin:$PATH"
     eval "$(pyenv init -)"
-    pyenv install pypy-4.0.1
-    pyenv global pypy-4.0.1
+    pyenv install pypy-5.3.1
+    pyenv global pypy-5.3.1
 fi
 
 pip install -U tox coveralls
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -46,8 +46,8 @@
             "upgrade PyPy to use this library."
         )
 else:
-    requirements.append("cffi>=1.1.0")
-    setup_requirements.append("cffi>=1.1.0")
+    requirements.append("cffi>=1.7.0")
```
reaperhulk: Bumping the cffi requirement to require the latest without a reason (e.g. we need a feature only provided by a newer version) is not something we want to do.

reaperhulk:
If you remove that cffi bump we can merge the pypy upgrade though!

setup.py
```diff
@@ -46,8 +46,8 @@
             "upgrade PyPy to use this library."
         )
 else:
-    requirements.append("cffi>=1.7.0")
-    setup_requirements.append("cffi>=1.7.0")
+    requirements.append("cffi>=1.4.1")
+    setup_requirements.append("cffi>=1.4.1")
```
lmctv: I think this could help in correctly solving #186 insetead of working around the problem like in #187

setup.py
```diff
@@ -46,8 +46,8 @@
             "upgrade PyPy to use this library."
         )
 else:
-    requirements.append("cffi>=1.1.0")
-    setup_requirements.append("cffi>=1.1.0")
+    requirements.append("cffi>=1.4.1")
```
reaperhulk: Why the increase to 1.4.1?

lmctv:
OK: I removed the cffi minimum version requirement and will put the new requirement when I come to the init_once PR.

PR #199, opened 2016-08-19
Use init_once around library initialization

lmctv:
Making for a race free version of #187 @asn-the-goblin-slayer,  @JackWink

### DIFF
--- a/src/nacl/bindings/sodium_core.py
+++ b/src/nacl/bindings/sodium_core.py
@@ -13,8 +13,13 @@
 # limitations under the License.
 from __future__ import absolute_import, division, print_function
 
-from nacl._sodium import lib
+from nacl._sodium import ffi, lib
 from nacl.exceptions import CryptoError
+
+
+def _sodium_init():
+    if lib.sodium_init() == -1:
+        raise CryptoError("Could not initialize sodium")
 
 
 def sodium_init():
@@ -22,5 +27,4 @@ def sodium_init():
     Initializes sodium, picking the best implementations available for this
     machine.
     """
-    if lib.sodium_init() != 0:
-        raise CryptoError("Could not initialize sodium")
+    ffi.init_once(_sodium_init, "libsodium")
--- a/setup.py
+++ b/setup.py
@@ -46,8 +46,8 @@ if platform.python_implementation() == "PyPy":
             "upgrade PyPy to use this library."
         )
 else:
-    requirements.append("cffi>=1.1.0")
-    setup_requirements.append("cffi>=1.1.0")
+    requirements.append("cffi>=1.4.1")
+    setup_requirements.append("cffi>=1.4.1")
 
 
 def here(*paths):


src/nacl/bindings/sodium_core.py
```diff
@@ -13,14 +13,18 @@
...
 from __future__ import absolute_import, division, print_function
 
-from nacl._sodium import lib
+from nacl._sodium import ffi, lib
 from nacl.exceptions import CryptoError
 
 
-def sodium_init():
+def _sodium_init():
     """
```
dstufft: Docstring should be on public function.

lmctv: Sorry! Updated PR in flight.

asn-the-goblin-slayer:
Thx for fixing this!

JackWink:
I can't remember if PyNaCl allows you to bring your own libsodium or if it uses a bundled version, but libsodium has made `sodium_init` thread safe starting in version 1.0.11. 

This seems like an alright fix for older versions of libsodium, though one possible issue is that you are no longer able to call `sodium_init` more than once. That said, I'm not sure if there is a legitimate case where you'd want to call `sodium_init` multiple times.

lmctv:
In as much as I can see, sodium_init itself is written to avoid repeated initializations.  The ffi.init_once call would still avoid useless crossing of the c/python boundary.

From what I see the boundled version is still 1.0.8 which predates the 1.0.11 thread safety enhancements.

Thank you very much.

Il 19 agosto 2016 21:58:03 CEST, Jack Wink <notifications@github.com> ha scritto:
>I can't remember if PyNaCl allows you to bring your own libsodium or if
>it uses a bundled version, but libsodium has made `sodium_init` thread
>safe starting in version 1.0.11. 
>
>This seems like an alright fix for older versions of libsodium, though
>one possible issue is that you are no longer able to call `sodium_init`
>more than once. That said, I'm not sure if there is a legitimate case
>where you'd want to call `sodium_init` multiple times.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/pyca/pynacl/pull/199#issuecomment-241119519

+-------------------------+----------------------------------------------+
|   Lorenzo M.  Catucci   | Centro di Calcolo e Documentazione           |
| catucci@ccd.uniroma2.it | Università degli Studi di Roma "Tor Vergata" |
|                         | Via O. Raimondo 18 ** I-00173 ROMA  ** ITALY |
|  Tel. +39 06 7259 2255  | Fax. +39 06 7259 2125                        |
+-------------------------+----------------------------------------------+

JackWink:
Agreed, I think this is an acceptable fix. It looks like PyNaCl supports bringing your own libsodium, so even if the bundled version was updated, we would still want to handle thread safety from the python side for older system distributions.

lmctv:
@warner  would you mind taking a look at this PR? 633ece35ac0df3e3f2deb656b1f121d97af3e5d3 is needed also for all of the other ones, since they use ffi.memmove, which has been implemented in 1.3,

PR #204, opened 2016-09-19
@malthe @owockEznan Add crypto pwhash scryptsalsa208sha256 updated

lmctv:
#204 This is just a little more than updating #130 , rebasing on master and adding support for memory and operations limit tweaks to the high level interfaces.

@malthe , @owockEznan


lmctv:
@warner this is low-priority for my needs, it just seemed like a very low hanging fruit.

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if len(passwd_hash) != STRBYTES - 1:
+        raise ValueError("Invalid password hash")
+
+    if lib.crypto_pwhash_scryptsalsa208sha256_str_verify(passwd_hash,
+                                                         passwd,
+                                                         len(passwd)) == 0:
+        return True
+    else:
+        return False
```
reaperhulk: I'd prefer an exception rather than `False` here. Is there another area in pynacl that does this already?

tests/test_pw_hash.py
```diff
@@ -0,0 +1,147 @@
...
+                          "opslimit", "memlimit",
+                          "expected"), [
+        (
+            32,
+            b"The quick brown fox jumps over the lazy dog.",
+            b"ef537f25c895bfa782526529a9b63d97",
+            20000,
+            (2 ** 20) * 100,
+            (b"\x10e>\xc8A8\x11\xde\x07\xf1\x0f\x98"
+             b"EG\xe6}V]\xd4yN\xae\xd3P\x87yP\x1b\xc7+n*")
```
reaperhulk: Source of this test vector?

tests/test_pw_hash.py
```diff
@@ -0,0 +1,147 @@
...
+            (2 ** 20) * 100
+        ),
+    ],
+)
+def test_bindings_wrong_salt_length(size, password, salt,
+                                    opslimit, memlimit):
+    with pytest.raises(ValueError):
+        res = crypto_pwhash_scryptsalsa208sha256(size, password, salt,
+                                                 opslimit, memlimit)
+        assert res is not None
```
reaperhulk: This line will never be executed.

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
```diff
@@ -0,0 +1,134 @@
...
+    "SALTBYTES",
+    "STRBYTES",
+    "OPSLIMIT_INTERACTIVE",
+    "MEMLIMIT_INTERACTIVE",
+    "OPSLIMIT_SENSITIVE",
+    "MEMLIMIT_SENSITIVE"
+]
+
+
+SALTBYTES = lib.crypto_pwhash_scryptsalsa208sha256_saltbytes()
```
lmctv: Will rename, even though I think preserving the c "namespacing" is redundant, since modules are proper namespaces  on their own...

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
```diff
@@ -0,0 +1,134 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import, division, print_function
+
+from nacl._sodium import ffi, lib
+from nacl.exceptions import CryptoError
+
+__all__ = [
```
lmctv: Will remove. Inherited from Malthe Borch's 3ff92c0c

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
lmctv: Inherited from owockEznan d819fbcf; while I can see the merits of both options, I've always considered a password checker a (strange) form of equality checker.
As a matter of facts, I just went to passlib's documentation at https://passlib.readthedocs.io/en/stable/ and I see on the very first page an example of an has verifier returning True/False, instead of raising an exception.

tests/test_pw_hash.py
lmctv: Yes, I just forgot removing the assignment and the useless assert.

tests/test_pw_hash.py
lmctv: Inherited from Malthe Borch's d889f85 .

Just double checked with cpython3.6 hashlib implementation, by manually running libsodium's 
   pickparams(...) and finding out that with opslimit=20000 and memlimit=(2 ** 20) * 100 
n = 128, r = 8, p =1
```
>>> hashlib.scrypt(b'The quick brown fox jumps over the lazy dog.',
                               salt=b"ef537f25c895bfa782526529a9b63d97",
                               n=1024, r=8, p=1, maxmem = (2**20)*100, dklen=32)
b'\x10e>\xc8A8\x11\xde\x07\xf1\x0f\x98EG\xe6}V]\xd4yN\xae\xd3P\x87yP\x1b\xc7+n*'
>>>
```

lmctv:
I just came to the conclusion the right function to bind is the low-level one, which is API compatible with the one in cpython3.6's hashlib . As things stand, I'll have to:
 - bind pickparams and crypto_pwhash_scryptsalsa208sha256_ll
 - remove the bindings for the simplified crypto_pwhash_scryptsalsa208sha256
 - reimplement the python shim for  crypto_pwhash_scryptsalsa208sha256 in terms of _ll and pickparams to keep the simplified libsodium-like API working.

Afterwards, I think I'd put an hashlib compatible shim for _ll into hashlib.

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
```diff
@@ -0,0 +1,134 @@
...
+    lib.crypto_pwhash_scryptsalsa208sha256_opslimit_sensitive()
+MEMLIMIT_SENSITIVE = \
+    lib.crypto_pwhash_scryptsalsa208sha256_memlimit_sensitive()
+
+
+def crypto_pwhash_scryptsalsa208sha256(outlen, passwd, salt,
+                                       opslimit=OPSLIMIT_SENSITIVE,
+                                       memlimit=MEMLIMIT_SENSITIVE
+                                       ):
+    """
```
lmctv: I was forgetting: another reason why I've defined the short local names was trying to get along with flake8 both in the function definitions and in the docstrings.

lmctv:
I've just pushed a commit in the hope the PR now fits better with pynacl's goals and conventions.

Since the pickparams function is not exported, I'm unsure if a python reimplementation would make sense. For this reason, I'm deleting my previous comment about replacing `crypto_pwhash_scryptsalsa208sha256` with the "low level" alternative.

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
```diff
@@ -0,0 +1,137 @@
...
+
+    :param outlen: int
+    :param passwd: bytes
+    :param salt: bytes  *must* be *exactly* :py:const:`.SALTBYTES` long
+    :param opslimit: int
+    :param memlimit: int
+    :rtype: bytes
+    """
+
+    if len(salt) != SALTBYTES:
```
lmctv: I think this would be an excellent fit for the utils.ensure() assert replacer from #223

reaperhulk:
Looks like there are some branch misses for raising CryptoError based on codecov.

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
```diff
@@ -0,0 +1,142 @@
...
+        raise ValueError("Invalid salt")
+
+    buf = ffi.new("unsigned char[]", outlen)
+
+    ret = lib.crypto_pwhash_scryptsalsa208sha256(buf, outlen, passwd,
+                                                 len(passwd), salt,
+                                                 opslimit, memlimit)
+
+    if ret != 0:
+        raise CryptoError("Key derivation fails!")
```
lmctv: Hopefully unreachable, since scryptsalsa should never return a non-zero result. #223 could be useful both here and at line 119

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
reaperhulk: I just talked with @warner and we'd like return True/raise exception on false. I know this isn't what passlib does but we both prefer failure via exception (and other pyca projects like cryptography take this approach).

lmctv:
OK, with d725d5d the password checker returns True on match and raises a ValueError in mismatch,

src/nacl/bindings/crypto_pwhash_scryptsalsa208sha256.py
```diff
@@ -0,0 +1,145 @@
...
+
+    ensure(len(passwd_hash) == STRBYTES - 1, 'Invalid password hash',
+           raising=exc.ValueError)
+
+    ret = lib.crypto_pwhash_scryptsalsa208sha256_str_verify(passwd_hash,
+                                                            passwd,
+                                                            len(passwd))
+    ensure(ret == 0,
+           "Wrong password",
+           raising=exc.ValueError)
```
reaperhulk: I think this should be a separate exception like InvalidKey rather than the same exception you'd get from a wrong length.

lmctv: OK, will define a specific exception. While at it, I don't like the asimmetry between an expected result on match and an exception on mismatch; if we choose the exception route, I think we shouldn't expose a return value.

lmctv: Specific exception defined.

warner:
@reaperhulk and I were just looking at this. It appears that libsodium (and this patch) is emphasizing an odd flavor of scrypt functionality, namely the form that takes an `opslimit` and `memlimit`, then figures out what N/r/p parameters would meet that requirement, generates a random salt, and then produces the hash.

We're thinking that it'd be better to expose the real `scrypt(key, salt, N, r, p, outlen)` function (i.e. `crypto_pwhash_scryptsalsa208sha256_ll()`), since that's what's going to enable interoperability with any other application that uses scrypt. The opslimit/memlimit thing is pretty specialized for the original tarsnap password-based file-encryption use case (which, if I remember correctly, was even more specialized: you gave it a limit denominated in *seconds*, and it ran some speed tests to figure out the maximum stretch it could do by the deadline).

That function should be equivalent to the `hashlib.scrypt` in python3.6 (when compiled against openssl-1.1.0), or `cryptography.hazmat.primitives.kdf.scrypt.Scrypt`.

If we do that, I'm undecided whether it's useful to expose the `_verify()` form, or to have apps call `scrypt()` and one of the constant-time `compare()` functions.

lmctv:
@reaperhulk @warner: I would just as well bind only the _ll "variant", to get API compatibility with cpython 3.6;  the only reason to bind ``crypto_pwhash_scryptsalsa208sha256_str`` and the corresponding ``crypto_pwhash_scryptsalsa208sha256_str_verify`` is the fact they use the serialization format specified for modular crypt(3) at https://gitlab.com/jas/scrypt-unix-crypt/blob/master/unix-scrypt.txt which preserves the N, r, p parameters, and should enhance the interoperability of the serialized hash and obviate the need to define a serialization format to preserve the salt and the parameters.

```
>>>> nacl.bindings.crypto_pwhash_scryptsalsa208sha256_str('mypassword',
                                                                                                     opslimit=2000000,
                                                                                                     memlimit=32* (2**21))
'$7$D6..../....mMXMx7BkguY6uyDGkT7O6OBPPpC4IkirfwokjcRsVeB$bM.Ui2t5vRswaXJXttE57IIFPp88qMs54RUrryJbRO9'
```

lmctv:
@warner I've just started with _ll binding implementation. It still needs tests and the nacl.pwhash shim outside of bindings.
I'm still unsure if deleting the "simplified" variant, the _str one and the crypt(3) checker, but I'm leaning there...

reaperhulk:
I'm confused by the codecov data, seems like it may be looking at the wrong base. Would you mind rebasing/merging master to see if we can force it to fix itself?

lmctv:
No, no, codecov is right: I just started implementing _ll, but there is no test data.

lmctv:
Added some tests for libsodium's low-level implementation.
API level bindings still are still missing.
I'm unsure about the last test vector: should I leave it in, with the associated 2G maxmem, or keep it out?

src/bindings/crypto_pwhash_scryptsalsa208sha256.h
```diff
@@ -0,0 +1,49 @@
...
+
+
+size_t crypto_pwhash_scryptsalsa208sha256_saltbytes(void);
+size_t crypto_pwhash_scryptsalsa208sha256_strbytes(void);
+size_t crypto_pwhash_scryptsalsa208sha256_opslimit_interactive(void);
+size_t crypto_pwhash_scryptsalsa208sha256_memlimit_interactive(void);
+size_t crypto_pwhash_scryptsalsa208sha256_opslimit_sensitive(void);
+size_t crypto_pwhash_scryptsalsa208sha256_memlimit_sensitive(void);
+
+int crypto_pwhash_scryptsalsa208sha256(unsigned char * const out,
```
reaperhulk: Does this need to be bound any more?

lmctv: The are still needed for the _str variant.

lmctv: I was referring to the _limits bindings. Will send a cleanup PR to remove crypto_pwhash_scryptsalsa208sha256 laterali today.

reaperhulk: Okay, good enough for me. Merging 💃

PR #206, opened 2016-09-20
WIP: Add bindings for siphash-2-4

lmctv:
also known as shorthash in libsodium circles...


reaperhulk:
@lmctv you're fine asking for help/reviews in #cryptography-dev btw. cc'ing @warner to see what he thinks about this :)

warner:
Seems reasonable to me. I'm a bit confused about the difference between "shorthash" and "siphash", but it looks like libsodium is using `src/libsodium/crypto_shorthash/siphash24/*`, so following whatever they do is appropriate.

(which reminds me, we should circle back around to renaming pynacl to pysodium. I'm ok with making bindings to anything that's already in libsodium).

lmctv:
@warner sodium does "functional" renaming of the primitives, but still leaves thei original names available; I prefer keeping both the original and the renamed entry points for clarity.

src/nacl/bindings/crypto_shorthash.py
```diff
@@ -0,0 +1,38 @@
...
+
+    :param data:
+    :type data: bytes
+    :param key:
+    :type key: bytes
+    """
+    _key = ffi.new("unsigned char[]", KEYBYTES)
+    digest = ffi.new("unsigned char[]", BYTES)
+    ffi.memmove(_key, key, min(len(key), KEYBYTES))
+    # The key will be truncated or zero-padded to KEYBYTES length
```
reaperhulk: Is this a common convention? It seems odd to me to not just require that `len(key) == KEYBYTES`

src/nacl/bindings/crypto_shorthash.py
```diff
@@ -0,0 +1,38 @@
...
+    rc = lib.crypto_shorthash_siphash24(digest, data, len(data), _key)
+
+    assert rc == 0
```
reaperhulk: This is fine for now, but #212 applies for the future.

tests/test_shorthash.py
```diff
@@ -0,0 +1,130 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import, division, print_function
+
+import pytest
+
+from nacl.bindings import crypto_shorthash_siphash24
+
+HASHES = [
```
reaperhulk: Where do these come from? Can you add a comment explaining their provenance?

lmctv: Added, it's in 6459e3b .

src/nacl/bindings/crypto_shorthash.py
lmctv: I've just been following what I'd done for blake2b, whose key parameter is zero padded.

lmctv:
@reaperhulk  thank you very much for your first review. Let me know if you prefer this not-squashed commit sequence or a squashing rebase instead.

src/nacl/bindings/crypto_shorthash.py
```diff
@@ -0,0 +1,37 @@
...
+def crypto_shorthash_siphash24(data, key):
+    """Compute a fast, cryptographic quality, keyed hash of the input data
+
+    :param data:
+    :type data: bytes
+    :param key: len(key) must be equal to
+                :py:data:`.KEYBYTES` (16)
+    :type key: bytes
+    """
+    assert len(key) == KEYBYTES
```
reaperhulk: Let's make this `if len(key) != KEYBTES: raise ValueError("Key length must be {0} bytes".format(KEYBYTES))` or something like that. Then we can update the test to look for ValueError

lmctv: Changed as requested. Thank you once more.

Issue #21, opened 2013-03-17
Random Nonces by Default?


dstufft:
There are a number of places where the api expects a nonce. Ideally a nonce should never be reused. However DJB has stated for both Symmetrical and Asymmetrical encryption that with a random nonce the collision risk is negligible[1][2].

Currently PyNaCl documentation has lots of warnings but given with the "Crypto without the sharp edges" ideal should PyNaCl by default generate a random nonce? This will need a resolution to https://github.com/dstufft/pynacl/issues/20 or will need to change the `encrypt()` function to return a tuple of `(nonce, ciphertext)`.

[1] http://nacl.cr.yp.to/box.html
[2] http://nacl.cr.yp.to/secretbox.html

warner:
Oh, while writing one app, I found it'd be handy to provide the nonce as an integer (encrypting multiple segments of a large file, holding only one segment in memory at a time, using the nonce as the segment number). So maybe have `encrypt()` accept a `nonce_int=` kwarg too (and obviously it's an error to pass both `nonce=` and `nonce_int=`). This would imply a particular encoding for the integer (little-endian vs big-endian) that isn't present in nacl itself. Does libsodium provide/endorse anything here?

I also thought it'd be useful to provide a utility function that would extract the nonce from a boxed or secretboxed message, and maybe a second one that turns it back into an integer. This would be handy for protocols that are rejecting duplicate/old/out-of-order messages: check the nonce before decrypting anything. I can't decide if this function should be an attribute of the key objects, or if they should be non-method functions in `public.py`/`secret.py`.. I guess I lean towards the latter.

lvh:
I didn't find anything endorsed while writing caesium, but went with big-endian: https://github.com/lvh/caesium/blob/master/src/caesium/crypto/secretbox.clj#L29-L42

reaperhulk:
This is done (it will be in 1.1.0). See #210

lvh:
This is fantastic and one of the best UX improvements I've seen in cryptographic software in a long time. Fantastic job @reaperhulk :)

PR #210, opened 2016-11-20
Generate random nonces if omitted in encryption calls

felipedau:
Hello!

I agree with @dstufft on #21 that nonces could be generated by default, as well as with @warner on the first solution he mentioned:

> One option we discussed on IRC was:
>
> - use a random nonce unless you call `encypt()` with a `nonce=` kwarg

How do you think this could be tested? I wrote one test for each method that just encrypts, decrypts and compares with the expected plaintext. Another test could work similarly to `test_random_bytes_produces_different_bytes` from `test_utils.py`, encrypting twice and making sure the nonces are different.

Could this be useful?

Thanks!

### DIFF
--- a/src/nacl/public.py
+++ b/src/nacl/public.py
@@ -139,10 +139,11 @@ class Box(encoding.Encodable, StringFixer, object):
 
         return box
 
-    def encrypt(self, plaintext, nonce, encoder=encoding.RawEncoder):
+    def encrypt(self, plaintext, nonce=None, encoder=encoding.RawEncoder):
         """
-        Encrypts the plaintext message using the given `nonce` and returns
-        the ciphertext encoded with the encoder.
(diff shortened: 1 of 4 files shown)


felipedau:
Although I've made multiple commits and new ones could also be made as the PR is discussed, I'd like to squash them into one single commit before merging (if this is accepted) as these changes are small and simple.

reaperhulk:
Note to self: Put in a separate PR to update the docs in `docs/`

felipedau:
Thanks @reaperhulk!

Let me know if you need help with the docs. Would examples be updated?

reaperhulk:
@felipedau if you'd like to update the docs that'd be great! Either update the examples or (perhaps more usefully) add a new example that shows what it looks like when you don't pass a nonce.

PR #201, opened 2016-09-03
Add blake2b/generichash bindings

lmctv:
The PR implements both the one-shot API and a hashlib like one completely fulfilling #178 

I experienced test failures stemming from missing python2.6 and python3.3 in newly created environments; since I've split out in #200 the installation of those missing interpreters, failures in testing can be expected.

PR #216, opened 2016-12-13
Add blake2/generichash oneshot implementation

lmctv:
formerly in #201

### DIFF
--- a/src/nacl/bindings/crypto_generichash.py
+++ b/src/nacl/bindings/crypto_generichash.py
@@ -1,0 +1,114 @@
+# Copyright 2013 Donald Stufft and individual contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import, division, print_function
+
+from six import integer_types
+
+from nacl import exceptions as exc
+
+from nacl._sodium import ffi, lib
+from nacl.utils import ensure
+
+crypto_generichash_BYTES = lib.crypto_generichash_blake2b_bytes()
+crypto_generichash_BYTES_MIN = lib.crypto_generichash_blake2b_bytes_min()
+crypto_generichash_BYTES_MAX = lib.crypto_generichash_blake2b_bytes_max()
+crypto_generichash_KEYBYTES = lib.crypto_generichash_blake2b_keybytes()
+crypto_generichash_KEYBYTES_MIN = lib.crypto_generichash_blake2b_keybytes_min()
+crypto_generichash_KEYBYTES_MAX = lib.crypto_generichash_blake2b_keybytes_max()
+crypto_generichash_SALTBYTES = lib.crypto_generichash_blake2b_saltbytes()
+crypto_generichash_PERSONALBYTES = \
+    lib.crypto_generichash_blake2b_personalbytes()
+
+_OVERLONG = '{0} length greater than {1} bytes'
+_TOOBIG = '{0} greater than {1}'
+
+
+def generichash_blake2b_salt_personal(data,
+                                      digest_size=crypto_generichash_BYTES,
+                                      key=b'', salt=b'', person=b''):
+    """One shot hash interface
+    :param data: the input data to the hash function
+    :param digest_size: must be at most
+                        :py:data:`.crypto_generichash_BYTES_MAX`;
+                        the default digest size is
+                        :py:data:`.crypto_generichash_BYTES`
+    :type digest_size: int
+    :param key: must be at most
+                :py:data:`.crypto_generichash_KEYBYTES_MAX` long
+    :type key: bytes
+    :param salt: must be at most
+                 :py:data:`.crypto_generichash_SALTBYTES` long;
+                 will be zero-padded if needed
+    :type salt: bytes
+    :param person: must be at most
+                   :py:data:`.crypto_generichash_PERSONALBYTES` long:
+                                          will be zero-padded if needed
+    :type person: bytes
+    :return: digest_size long digest
+    :rtype: bytes
+    """
+    ensure(isinstance(data, bytes),
+           'Input data must be a bytes sequence',
+           raising=exc.TypeError)
+
+    ensure(isinstance(key, bytes),
+           'Key must be a bytes sequence',
+           raising=exc.TypeError)
+
+    ensure(isinstance(salt, bytes),
+           'Salt must be a bytes sequence',
+           raising=exc.TypeError)
+
+    ensure(isinstance(person, bytes),
+           'Person must be a bytes sequence',
+           raising=exc.TypeError)
+
+    ensure(isinstance(digest_size, integer_types),
+           'Digest size must be an integer number',
+           raising=exc.TypeError)
+
+    ensure(digest_size <= crypto_generichash_BYTES_MAX,
+           _TOOBIG.format("Digest_size", crypto_generichash_BYTES_MAX),
+           raising=exc.ValueError)
+
+    ensure(len(key) <= crypto_generichash_KEYBYTES_MAX,
+           _OVERLONG.format("Key", crypto_generichash_KEYBYTES_MAX),
+           raising=exc.ValueError)
+
+    ensure(len(salt) <= crypto_generichash_SALTBYTES,
+           _OVERLONG.format("Salt", crypto_generichash_SALTBYTES),
+           raising=exc.ValueError)
+
+    ensure(len(person) <= crypto_generichash_PERSONALBYTES,
+           _OVERLONG.format("Person", crypto_generichash_PERSONALBYTES),
+           raising=exc.ValueError)
+
+    digest = ffi.new("unsigned char[]", digest_size)
+
+    # both _salt and _personal must be zero-padded to the correct length
+    _salt = ffi.new("unsigned char []", crypto_generichash_SALTBYTES)
+    _person = ffi.new("unsigned char []", crypto_generichash_PERSONALBYTES)
+
+    ffi.memmove(_salt, salt, len(salt))
+    ffi.memmove(_person, person, len(person))
+
+    rc = lib.crypto_generichash_blake2b_salt_personal(digest, digest_size,
+                                                      data, len(data),
+                                                      key, len(key),
+                                                      _salt, _person)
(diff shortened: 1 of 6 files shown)


src/nacl/bindings/crypto_generichash.py
```diff
@@ -0,0 +1,50 @@
...
+    _salt = ffi.new("char []", crypto_generichash_SALTBYTES)
+    _personal = ffi.new("char []", crypto_generichash_PERSONALBYTES)
+
+    ffi.memmove(_salt, salt, min(len(salt), crypto_generichash_SALTBYTES))
+    ffi.memmove(_personal, personal,
+                min(len(personal), crypto_generichash_PERSONALBYTES)
+                )
+    # _salt and _personal will be truncated or zero-padded
+    # to the correct length
+
```
lmctv: Both parameters are defined as being of a fixed size and zero-filled.
This is the reason why I think the memmove usage to be correct.

lmctv:
@reaperhulk  I've changed my implementation so that it now mimics the behaviour of blake2b as implemented in python 3.6 by just zero padding, and raising an exception for too long salt and personal.

zookozcash:
I also read through these patches at the same time as https://github.com/pyca/pynacl/pull/217#issuecomment-266923156

lmctv:
@reaperhulk  I've splitted the PR into three parts, in the hope this will help reviews.

reaperhulk:
This can be rebased now

lmctv:
@reaperhulk rebased, as requested.

src/nacl/bindings/crypto_generichash.py
```diff
@@ -0,0 +1,72 @@
...
+    lib.crypto_generichash_blake2b_personalbytes()
+
+_OVERLONG = '{0} length greater than {1} bytes'
+_TOOBIG = '{0} greater than {1}'
+
+
+def generichash_blake2b_salt_personal(data,
+                                      digest_size=crypto_generichash_BYTES,
+                                      key=b'', salt=b'', person=b''):
+    """One time hash interface"""
```
warner: might be nice if the docstring said "one *shot* hash interface", instead of one-time

src/nacl/hash.py
```diff
@@ -17,10 +17,80 @@
...
 
 def sha256(message, encoder=nacl.encoding.HexEncoder):
     return encoder.encode(nacl.bindings.crypto_hash_sha256(message))
 
 
 def sha512(message, encoder=nacl.encoding.HexEncoder):
     return encoder.encode(nacl.bindings.crypto_hash_sha512(message))
+
+
+def generichash(data, digest_size=BYTES, key=b'',
```
warner: I'm dubious about whether such a generic-sounding name as `generichash` should be exposed. I know nacl likes to offer a short(er) name for the specific recommended algorithm for each class of operations, but I don't think hiding the specific hash being used serves anybody very well. Maybe we should leave this function out, and just offer `blake2b()` instead?

src/nacl/hash.py
```diff
@@ -17,10 +17,80 @@
...
+                        the default length is :py:data:`.BYTES`
+    :type digest_size: int
+    :param key: the key to be set for keyed MAC/PRF usage; if set, key size
+                should be comprised between :py:data:`.KEYBYTES_MIN`
+                and :py:data:`.KEYBYTES_MAX`
+    :type key: bytes
+    :param salt: an initialization salt of lenght :py:data:`.SALTBYTES`.
+                 It will be zero-padded or truncated if needed.
+    :type salt: bytes
+    :param person: a personalization string of lenght
```
warner: Any thoughts about calling this `personalization=` instead of `person=`? The latter sounds pretty weird.

reaperhulk:
Looks like we need a new `hash.rst` in `docs` documenting this (see the public/secret/signing docs for a sample of what we want it to look like)

src/nacl/bindings/crypto_generichash.py
lmctv: I think I wrote the docstring before thinking about the "one shot" phrase... Will change.

src/nacl/hash.py
lmctv: While for my own expected use cases I only need the salted and personalized version, I think feature-parity of the bindings with the C library can aid those coming to the python bindings from libsodium's documentation, which refers to the blake2 name (and API) only in a page about key derivation.

src/nacl/hash.py
lmctv: I tried to implement the same call signature used in python3.6's stdlib version, modulo missing support for tree hashes in libsodium's version.

src/nacl/bindings/crypto_generichash.py
```diff
@@ -0,0 +1,72 @@
...
+
+    ensure(len(person) <= crypto_generichash_PERSONALBYTES,
+           _OVERLONG.format("Person", crypto_generichash_PERSONALBYTES),
+           raising=exc.ValueError)
+
+    # both _salt and _personal must be zero-padded to the correct length
+    _salt = ffi.new("char []", crypto_generichash_SALTBYTES)
+    _person = ffi.new("char []", crypto_generichash_PERSONALBYTES)
+
+    ffi.memmove(_salt, salt, len(salt))
```
reaperhulk: Let's not use `memmove` here. Users should pass the full length required or `None`. If they pass `None` we should pass `ffi.NULL` to the function rather than building the null byte string on our end (this goes for key, salt, and personalization).

lmctv: For all the parameters key, salt, personal, I see the very same pattern in cpython implementation: respectively at lines 130, 141 and 228 of
https://github.com/python/cpython/blob/c30098c8c6014f3340a369a31df9c74bdbacc269/Modules/_blake2/blake2b_impl.c .

Here are a couple of test from cpython3.6's hashlib, where I've tried putting in short salt and personal, and the zero-padded version of the same:
```
>>> hashlib.blake2b(b'asdf', key=b'asd', salt=b'sdf', person=b'ply').hexdigest()
'931a7d96acc24a7954420d45404e5351465e00c45188341da66c0fe01254c8e5a65a320a2f399006afe56b60262fe918dff9ef5cd62eb7850b2f09d5bb12e364'
>>> hashlib.blake2b(b'asdf', key=b'asd', salt=b'sdf'+13*b'\x00', person=b'ply').hexdigest()
'931a7d96acc24a7954420d45404e5351465e00c45188341da66c0fe01254c8e5a65a320a2f399006afe56b60262fe918dff9ef5cd62eb7850b2f09d5bb12e364'
>>> hashlib.blake2b(b'asdf', key=b'asd', salt=b'sdf'+13*b'\x00', person=b'ply'+13*b'\x00').hexdigest()
'931a7d96acc24a7954420d45404e5351465e00c45188341da66c0fe01254c8e5a65a320a2f399006afe56b60262fe918dff9ef5cd62eb7850b2f09d5bb12e364'
```

lmctv: I've just seen it should be possible avoiding the ``memmove`` for the key, since it's block should be zero-padded on input.

lmctv: I didn't remember, but I've had removed the memmove in the first iterations of this PR.
The other two are operationally correct, and needed to avoid crashes under pypy.

lmctv:
There has been a connection failure while executing Job #560.11  https://travis-ci.org/pyca/pynacl/jobs/189626734 .
It would be useful if someone with admin access to travis could restart just the failed job.

lmctv:
Travis' network hates me!  In build job #565.2 -  https://travis-ci.org/pyca/pynacl/jobs/189742687  - while

```
Downloading virtualenv-15.1.0-py2.py3-none-any.whl (1.8MB)
    45% |██████████████▋                 | 829kB 10.6MB/s eta 0:00:01Exception:
```

there has been a disconnection:
```
   [...]
  File "/home/travis/.pyenv/versions/3.3.6/lib/python3.3/ssl.py", line 352, in read
    v = self._sslobj.read(len, buffer)
ConnectionResetError: [Errno 104] Connection reset by peer
```

lmctv:
@warner I've added a proper docstring for the low-level python bindings, and corrected that "one time" in "one shot" as suggested; as for nacl.hash.generichash , now it's simply another name for nacl.hash.blake2b instead of a mostly copy-pasted implementation.

Could you do another review pass?

lmctv:
@warner , @reaperhulk  I've added a little bit of documentation, as requested by Paul, and I think I've already complied with Brian's requests last week. Would you mind one more review pass?

tests/test_generichash.py
```diff
@@ -0,0 +1,109 @@
...
+    (b'salt', b'key', 17*b'\xaa', 16*b'\x55', 64, b'will raise'),
+    (b'personal', b'key', 16*b'\xaa', 17*b'\x55', 64, b'will raise'),
+    (b'digest_size', b'key', 16*b'\xaa', 16*b'\x55', 65, b'will raise'),
+]
+
+
+def _read_crypto_test_vectors(fname):
+    vectors = []
+    path = os.path.join(os.path.dirname(__file__), "data", fname)
+    with open(path, "r") as fp:
```
reaperhulk: Open this with `rb` and then you shouldn't need the `bytes_as_string` helper since on py3 the resulting data will be a bytestring as expected.

lmctv: I've just tried to implement your suggestion, but, while all goes well on python2.7, all I've got on python3.5 is a crash, deep inside py.test during test collection, before the real test could even start:

```
==================================== ERRORS ====================================
__________________ ERROR collecting tests/test_generichash.py __________________
venv-35/lib/python3.5/site-packages/_pytest/runner.py:163: in __init__
    self.result = func()
venv-35/lib/python3.5/site-packages/_pytest/main.py:464: in _memocollect
    return self._memoizedcall('_collected', lambda: list(self.collect()))
venv-35/lib/python3.5/site-packages/_pytest/main.py:335: in _memoizedcall
    res = function()
# # # [  .  .  .  ]
    _MultiCall(methods, kwargs, hook.spec_opts).execute()
venv-35/lib/python3.5/site-packages/_pytest/vendored_packages/pluggy.py:614: in execute
    res = hook_impl.function(*args)
venv-35/lib/python3.5/site-packages/_pytest/python.py:110: in pytest_generate_tests
    metafunc.parametrize(*marker.args, **marker.kwargs)
venv-35/lib/python3.5/site-packages/_pytest/python.py:856: in parametrize
    assert len(valset) == len(argnames)
E   AssertionError
```

I think it's better to leave things as they stand now.

lmctv: OK, I've managed to tame python3.5 bytes comparison, and therefore added a commit removing `bytes_as_string` usage as you requested.

I'm unsure if doing a more drastic rebase to remove the helper's definition from history and having it moved to the init/update/finalize PR, or leaving the commit in for (near?) future use.

reaperhulk: You can leave it here and we can discuss whether it needs to exist in the init/update/finalize PR :)

reaperhulk:
Since the next PR will immediately address the function that doesn't have coverage I'm merging.

lmctv:
@reaperhulk @warner  thank you very much for your help!

PR #217, opened 2016-12-13
Add blake2b/generichash init-update-finalize implementation

lmctv:
formerly in #201 Any suggestion pointing to a better naming is VERY well accepted

### DIFF
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -13,8 +13,8 @@ building blocks of other kind of algorithms and data structures.
 All of the hash functions exposed in :py:mod:`nacl.hash` can be used
 as data integrity checkers.
 
-Integrity check examples:
--------------------------
+Integrity check examples
+------------------------
 
 Message's creator perspective (:py:func:`~nacl.hash.sha256`,
                                :py:func:`~nacl.hash.sha512`,
@@ -85,8 +85,8 @@ Message's user perspective (:py:func:`~nacl.hash.sha256`,
         print(MSG.format(chk[0], eq_chk(dgst, chk[1])))
 
 
-Additional hashing usages for :class:`~nacl.hash.blake2b`:
-==========================================================
+Additional hashing usages for :class:`~nacl.hash.blake2b`
+=========================================================
 
 As already hinted above, traditional cryptographic hash functions can be used
 as building blocks for other uses, tipically combining a secret-key with
@@ -97,14 +97,22 @@ for message authentication and key derivation, replacing the ``HMAC`` construct
 and the ``HKDF`` one by setting the additional parameters ``key``, ``salt``
 and ``person``.
 
-Message authentication:
------------------------
+Please note that **key stretching procedures** like ``HKDF`` or
+the one outlined in `Key derivation`_ are **not** suited to derive
+a *crytpographically-strong* key from a *low-entropy input* like a plain-text
+password or to compute a strong *long-term stored* hash used as password
+verifier. The :py:mod:`~nacl.pw_hash` module provides implementations of the
+``scrypt`` password hash, suitable both for key generation, an long-term
+password storage.
+
+Message authentication
+----------------------
 
 To authenticate a message, using a secret key, the blake2b function
 must be called as in the following example.
 
-Message authentication example:
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Message authentication example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. code-block:: python
 
@@ -114,11 +122,14 @@ Message authentication example:
     msg = 16*b'256 BytesMessage'
     msg2 = 16*b'256 bytesMessage'
 
-    auth_key = b''.join([bytearray((i,)) for i in range(64)])
-    # should really be a cryptographic quality random key, like
-    # the ones you could get with nacl.utils.random(size=64)
+    auth_key = nacl.utils.random(size=64)
+    # the simplest way to get a cryptographic quality auth_key
+    # is to generate it with a cryptographic quality
+    # random number generator
 
-    auth1_key = b''.join([bytearray((i,)) for i in range(64,128)])
+    auth1_key = nacl.utils.random(size=64)
+    # generate a different key, just to show the mac is changed
+    # both with changing messages and with changing keys
 
     mac0 = blake2b(msg, key=auth_key, encoder=nacl.encoding.HexEncoder)
     mac1 = blake2b(msg, key=auth1_key, encoder=nacl.encoding.HexEncoder)
@@ -128,14 +139,14 @@ Message authentication example:
         print('Mac{0} is: {1}.'.format(i, mac))
 
 
-Key derivation:
----------------
+Key derivation
+--------------
 
 The blake2b algorithm can replace a key derivation function by
 following the lines of:
 
-Key derivation example:
-~~~~~~~~~~~~~~~~~~~~~~~
+Key derivation example
+~~~~~~~~~~~~~~~~~~~~~~
 
 .. code-block:: python
 
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -99,10 +99,10 @@ and ``person``.
 
 Please note that **key stretching procedures** like ``HKDF`` or
 the one outlined in `Key derivation`_ are **not** suited to derive
-a *crytpographically-strong* key from a *low-entropy input* like a plain-text
+a *cryptographically-strong* key from a *low-entropy input* like a plain-text
 password or to compute a strong *long-term stored* hash used as password
 verifier. The :py:mod:`~nacl.pw_hash` module provides implementations of the
-``scrypt`` password hash, suitable both for key generation, an long-term
+``scrypt`` password hash, suitable both for key generation, and long-term
 password storage.
 
 Message authentication
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -1,0 +1,159 @@
+Hashing
+=======
+
+.. currentmodule:: nacl.hash
+
+Cryptographic secure hash functions are irreversible transforms
+of input data to a fixed length `digest`.
+
+The standard properties of a cryptographic hash make these functions useful
+both for standalone usage as data integrity checkers, as well as ``black-box``
+building blocks of other kind of algorithms and data structures.
+
+All of the hash functions exposed in :py:mod:`nacl.hash` can be used
+as data integrity checkers.
+
+Integrity check examples:
+-------------------------
+
+Message's creator perspective (:py:func:`~nacl.hash.sha256`,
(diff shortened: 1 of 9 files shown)


zookozcash:
Okay I read through these patches. There are some things I don't know about the nacl API so I couldn't tell for sure if this was right, such as how it handles the salt and personalization string being shorter than max. I also don't understand github well enough to navigate it and figure out if the handling of shorter-than-max salt and personalization strings was changed in both the init-update-digest API and also the one-shot API. I liked the way there were tests and test-vectors! ☺ I give it a +1.

reaperhulk:
Okay this is ready for a rebase

lmctv:
OK, rebased.

lmctv:
I think this is ready for review.

src/nacl/bindings/crypto_generichash.py
```diff
@@ -112,3 +119,101 @@ def generichash_blake2b_salt_personal(data,
...
+    :type name: bytes
+    :param data:
+    :type data: bytes
+    """
+
+    ensure(isinstance(data, bytes),
+           'Input data must be a bytes sequence',
+           raising=exc.TypeError)
+
+    _state = ffi.cast("struct crypto_generichash_blake2b_state *", statebuf)
```
reaperhulk: This cast should be unnecessary as you've already cast it in `init`

src/nacl/bindings/crypto_generichash.py
```diff
@@ -112,3 +119,101 @@ def generichash_blake2b_salt_personal(data,
...
+
+    ensure(rc == 0, 'Unexpected failure',
+           raising=exc.RuntimeError)
+    return ffi.buffer(_digest, digest_size)[:]
+
+
+def generichash_blake2b_state_copy(statebuf):
+    """Return a copy of the given blake2b hash state"""
+
+    _statebuf = ffi.new("unsigned char[]", crypto_generichash_STATEBYTES)
```
reaperhulk: Cast this after the memmove and then the `update` and `final` casts become unnecessary.

src/nacl/bindings/crypto_generichash.py
```diff
@@ -112,3 +119,101 @@ def generichash_blake2b_salt_personal(data,
...
+    :param statebuf:
+    :type statebuf: bytes
+    :param digest_size:
+    :type digest_size: int
+    :return: the blake2 digest of the passed-in data stream
+    :rtype: bytes
+    """
+
+    _digest = ffi.new("unsigned char[]", crypto_generichash_BYTES_MAX)
+    _state = ffi.cast("struct crypto_generichash_blake2b_state *", statebuf)
```
reaperhulk: Unneeded once `copy` casts.

src/nacl/bindings/crypto_generichash.py
lmctv: I just tried following your suggestion, but I quickly get a segmentation fault, which I don't fully understand whether to blame to a lifecycle problem with the underlying buffer, or to the different sizes of the buffer itself and the embedded struct.

The only way I've found to avoid the segmentation fault, and hiding the cast has been defining an helper class to keep the state as in:
```
class _blake2b_state(object):
    """Utility class to keep around the blake2b state"""

    def __init__(self, buf = None):
        self.statebuf = ffi.new("unsigned char[]",
                                crypto_generichash_STATEBYTES)
        if buf is not None:
            ffi.memmove(self.statebuf, buf, crypto_generichash_STATEBYTES)

    @property
    def b2state(self):
        return ffi.cast("struct crypto_generichash_blake2b_state *",
                        self.statebuf)
```

and express the operations in terms of that partially opaque class like in:

```
def generichash_blake2b_update(state, data):
    """ [ . . . ] """
    rc = lib.crypto_generichash_blake2b_update(state.b2state, data, len(data))

```
but I'm really unsure about the worth of such an indirection.

I've added the test commit to a branch in my forked repository; to browse the changes, you can open

https://github.com/lmctv/pynacl/compare/add-generichash-iuf...lmctv:add-generichash-iuf-hide-casts?expand=1

reaperhulk: Yikes. I definitely find libsodium's sizing to be weird here:

"The crypto_generichash_state structure length is either 357 or 361 bytes. 64-bytes alignment is recommended for performance, but not required. For dynamically allocated states, crypto_generichash_statebytes() returns the rounded up structure size, and should be prefered to sizeof()."

Is there any real drawback to just doing `ffi.new("struct crypto_generichash_blake2b_state *")` and not allocating those extra bytes to make it 64 byte aligned?

reaperhulk: Oops, I guess to be more clear, I think we should do that unless there's genuinely a significant performance delta.

lmctv: At first, I tried doing the `ffi.new("struct crypto_generichash_blake2b_state *")` you just suggested, but I later went by the "`ffi.new("unsigned char[]",crypto_generichash_STATEBYTES)` and then cast" dance, to avoid random crashes under pypy .

reaperhulk: Hmm, I'm definitely not comfortable with the uncertainty about what's going on here. Maybe we  should just write a state_new() and state_free(state *) function in C and call those from the Python layer? Then we can hide this nonsense.

lmctv: Any suggestion  / example to follow in adding a single C file and the associate object file  to setup.py?

lmctv: @reaperhulk After stumbling in a custom version of https://bitbucket.org/cffi/cffi/issues/191 I realized we are using the crypto_generichash_blake2b_state * as a opaque buffer, since we don't really get into its internal structure. With my latest commit, there are neither casts nor crashes, and there is no need for a C-side allocator/deallocator couple.  Would you mind starting another review pass?

docs/hashing.rst
```diff
@@ -0,0 +1,159 @@
...
+
+    import nacl.encoding
+    from nacl.hash import blake2b
+
+    msg = 16*b'256 BytesMessage'
+    msg2 = 16*b'256 bytesMessage'
+
+    auth_key = b''.join([bytearray((i,)) for i in range(64)])
+    # should really be a cryptographic quality random key, like
+    # the ones you could get with nacl.utils.random(size=64)
```
reaperhulk: Why not just have the example do this?

docs/hashing.rst
```diff
@@ -0,0 +1,159 @@
...
+    mac2 = blake2b(msg2, key=auth_key, encoder=nacl.encoding.HexEncoder)
+
+    for i, mac in enumerate((mac0, mac1, mac2)):
+        print('Mac{0} is: {1}.'.format(i, mac))
+
+
+Key derivation:
+---------------
+
+The blake2b algorithm can replace a key derivation function by
```
reaperhulk: Please add some text about this being not suitable for password based key derivation.

lmctv: @reaperhulk I think 4b8428e should cover both your review comments for now. I'm trying to write a page on password hashing, but I would rather open a separate pull request for that.

lmctv:
Usual networking problems between travis and pypi: 4b8428e has been already tested on my "private" add-generichash-rework branch: https://travis-ci.org/lmctv/pynacl/builds/195541344 (see https://travis-ci.org/lmctv/pynacl/jobs/195541345 for the "failed" python2.7 test)

docs/hashing.rst
```diff
@@ -0,0 +1,170 @@
...
+the message via some construct like the ``HMAC`` one.
+
+The :class:`~nacl.hash.blake2b` hash function can be used directly both
+for message authentication and key derivation, replacing the ``HMAC`` construct
+and the ``HKDF`` one by setting the additional parameters ``key``, ``salt``
+and ``person``.
+
+Please note that **key stretching procedures** like ``HKDF`` or
+the one outlined in `Key derivation`_ are **not** suited to derive
+a *crytpographically-strong* key from a *low-entropy input* like a plain-text
```
reaperhulk: typo, cryptographically

docs/hashing.rst
```diff
@@ -0,0 +1,170 @@
...
+password or to compute a strong *long-term stored* hash used as password
+verifier. The :py:mod:`~nacl.pw_hash` module provides implementations of the
+``scrypt`` password hash, suitable both for key generation, an long-term
```
reaperhulk: an should be and here

tests/test_generichash.py
```diff
@@ -108,3 +109,78 @@ def test_hash_blake2b(message, key, salt, person, outlen, output):
...
+def test_overlong_blake2b_iuf_params(message, key, salt, person,
+                                     outlen, output):
+    with pytest.raises(exc.ValueError):
+        nacl.hashlib.blake2b(message, digest_size=outlen,
+                             key=key, salt=salt, person=person)
+
+
+def test_blake2_descriptors_presence():
+    h = nacl.hashlib.blake2b()
+    h.name
```
reaperhulk: Could these be asserts on the values please? I know that it's not a very useful test, but just in case a refactor messes with those properties in the future.

lmctv: Corrected both typos.

reaperhulk: Looks like we still need the asserts in this test and the one after it?

lmctv: I'm thinking to just remove the `generic` wart from the hashlib module; especially if you are cool with moving the hashlib compatible implementation of scrypt here.

reaperhulk: The `generichash` name? Yes please.

docs/hashing.rst
```diff
@@ -0,0 +1,170 @@
...
+
+        print(MSG.format(chk[0], eq_chk(dgst, chk[1])))
+
+
+Additional hashing usages for :class:`~nacl.hash.blake2b`
+=========================================================
+
+As already hinted above, traditional cryptographic hash functions can be used
+as building blocks for other uses, tipically combining a secret-key with
+the message via some construct like the ``HMAC`` one.
```
reaperhulk: One last typo: typically

lmctv: Correction pushed. Thanks!

src/nacl/hashlib.py
```diff
@@ -0,0 +1,103 @@
...
+                     :py:attr:`.SALT_SIZE` long; it will be zero-padded
+                     if needed
+        :type salt: bytes
+        :param person: a personalization string at most
+                       :py:attr:`.PERSONAL_SIZE` long; it will be zero-padded
+                       if needed
+        :type person: bytes
+        """
+
+        self.state = _b2b_init(key=key, salt=salt, person=person,
```
reaperhulk: State should be private. (Sorry to keep finding one thing at a time! Hard to review large PRs without having that occur)

lmctv: You're right! Change pushed.

Issue #202, opened 2016-09-13
Drop 2.6 support


reaperhulk:
Looking at the big query data for PyNaCl there were 39185 downloads for this project in August. Here's the top 30 version breakdown: 

```
1   3.5.2   9615     
2   2.7.7   9523     
3   3.5.1   7702     
4   null    3134     
5   2.7.12  1350     
6   2.7.6   1161     
7   2.7.3   1119     
8   2.7.9   967  
9   2.7.5   862  
10  2.7.8   748  
11  2.7.10  422
12  2.7.11  397  
13  2.7 365  
14  3.5.0   358  
15  3.4.3   318  
16  3.4.2   289  
17  3.4.4   223  
18  3.3.5   127  
19  2.7.11+ 112  
20  3.5.2+  51   
21  3.5 47   
22  3.5.0+  44   
23  3.5.1+  43   
24  3.6.0a3 33   
25  3.4.3+  28   
26  3.4.5   24   
27  2.7.12+ 23   
28  3.6.0a4 19   
29  2.6.6   11   
30  3.4 9   
```

Based on this I'd like to propose dropping Python 2.6 from our list of supported versions. @warner do you have an opinion here since you have some projects that depend completely on pynacl?

alex:
Can you group this data by major.minor

warner:
My stuff is all 2.7+, so I'm cool with dropping 2.6.

dstufft:
👍

alex:
Counts aggregated:
`{'2.6': 11, '2.7': 17049, '3.3': 127, '3.4': 891, '3.5': 17860, '3.6': 52}`

I'm ok with with dropping 2.6

PR #219, opened 2016-12-18
drop 2.6 support, fix a broken image.

reaperhulk:


alex:
Do we have data supporting that it's ok to drop 2.6?

reaperhulk:
Yep, sorry, forgot to cross link #202.

alex:
👍  is there any 2.6 compat code that needs to be deleted?

reaperhulk:
I'm not aware of any. Coverage is a bit lower than it should be in general so I'll probably be submitting some PRs for that next week.

Issue #212, opened 2016-12-11
Stop using asserts


reaperhulk:
If pynacl is invoked in an environment where `python -O` is used all the asserts are ignored. We should switch the asserts to a function call.

reaperhulk:
Done!

PR #223, opened 2016-12-21
Asserts replacer

lmctv:
@reaperhulk  could this somewhat fit your ideas on #212 solution?

### DIFF
--- a/src/nacl/utils.py
+++ b/src/nacl/utils.py
@@ -60,7 +60,7 @@ def random(size=32):
     return nacl.bindings.randombytes(size)
 
 
-def check_condition(cond, *args, **kwds):
+def ensure(cond, *args, **kwds):
     """
     Return if a condition is true, otherwise raise a caller-configurable
     :py:class:`Exception`
--- a/src/nacl/utils.py
+++ b/src/nacl/utils.py
@@ -65,9 +65,9 @@ def ensure(cond, *args, **kwds):
     """
     Return if a condition is true, otherwise raise a caller-configurable
     :py:class:`Exception`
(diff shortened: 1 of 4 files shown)


lmctv:
As always, I'd be grateful for any naming suggestion.

As for the API, I've thought about both the natural alternatives
```
def check_condition(condition, raised, *args):
```
or
```
def check_condition(condition, *args):
```
But I'd prefer being able to have the failed test raise a custom exception, while being able to just specify the condition in an assert-like guise.

lmctv:
@reaperhulk  I think this PR and #216 are both ready for review.

lmctv:
Added a test for a custom exception as requested. While at it, I've corrected some other tests, and added a pynacl AssertionError, which can replace stdlib AssertionError whithout real changes to downstream expectations.

src/nacl/utils.py
```diff
@@ -58,3 +59,24 @@ def __str__(self):
 
 def random(size=32):
     return nacl.bindings.randombytes(size)
+
+
+def ensure(cond, *args, **kwds):
+    """
+    Return if a condition is true, otherwise raise a caller-configurable
+    :py:class:`Exception`
+    :param bool cond: the condition to be chacked
```
reaperhulk: checked

src/nacl/utils.py
```diff
@@ -58,3 +59,24 @@ def __str__(self):
...
+    :param sequence args: the arguments to be passed to the exception's
+                          consructor
```
reaperhulk: constructor

lmctv: Corrected both typos

PR #224, opened 2017-01-02
We do the py36

alex:


warner:
I think travis hasn't added py3.6 (-final) support yet (https://docs.travis-ci.com/user/languages/python/). I had to use `3.6-dev` in my `.travis.yml` for all my other projects. Hopefully now that 3.6 is out for real, Travis will update their environment soon.

Issue #185, opened 2016-04-27
Make "bundled" the default


oberstet:
Currently, PyNaCl will be default look for a system libsodium, and to use the bundled lib, one needs to have an environment variable `SODIUM_INSTALL=bundled` set.

This is IMO not the best, because:
1. it is not possible to set environment variables from wheel based installs (when X depends on PyNaCl, and X is a universal wheel, then X cannot be installed easily forcing bundled)
2. System libsodium is often outdated. This is a potential security issue. Also, I'd expect the bundled libsodium to be more thoroughly tested/compatible with the bundled libsodium rather than any system one)

Hence, I would like to propose making "bundled" the default, and only override that via `SODIUM_INSTALL=system`.

> This was discussed on IRC .. @dstufft

reaperhulk:
This is done (it will be in 1.1.0)

PR #226, opened 2017-01-05
don't try to detect libsodium, default to bundled

### LINKED ISSUE #184: Fallback to "bundled" when not both libsodium and headers are there
Currently, PyNaCl will check for availability of system libsodium, but it does not check for development headers. And then fails to build.

It would be nice if it would check for both system libsodium _and_ headers, and not both are available, automatically fall back to a "bundled" build.

> This was discussed on IRC .. @dstufft

reaperhulk:
fixes #184 and #185

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,9 +33,6 @@ from setuptools import Distribution, setup
 from setuptools.command.install import install
 
 
-SODIUM_MAJOR = 7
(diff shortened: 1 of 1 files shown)


lmctv:
Great!

warner:
yay!

oberstet:
Thanks for that!

PR #228, opened 2017-01-05
return the byte string as a str in py3

### LINKED ISSUE #205: SecretBox.__unicode__() attribute missing for Py3
Try to `print()` a SecretBox object and you get the following:

```
../python3.4/site-packages/nacl/utils.py", line 54, in __str__
    return self.__unicode__()
AttributeError: 'SecretBox' object has no attribute '__unicode__'
```

reaperhulk:
This encodes it a bit, but you shouldn't be printing things except for debug anyway.

Fixes #205

### DIFF
--- a/src/nacl/utils.py
+++ b/src/nacl/utils.py
@@ -51,7 +51,7 @@ class StringFixer(object):
 
     def __str__(self):
         if six.PY3:
-            return self.__unicode__()
+            return str(self.__bytes__())
         else:
             return self.__bytes__()
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -14,6 +14,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import nacl.secret
 import nacl.utils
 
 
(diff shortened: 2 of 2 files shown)


blag:
This looks perfectly cromulent to me.

lmctv:
@reaperhulk if we should keep the StringFixer mix-in, this LGTM.

reaperhulk:
I honestly don't remember the context for why I decided this was a better approach than stripping out that mixin... If you're comfortable with it then we can merge it and move forward. We can always revisit if necessary.

lmctv:
@reaperhulk I think It's better if we just fix this for the time being, an later decide when to insert a explicit deprecation leading to the removal of the mixin usage.

@reaperhulk @alex how should I do the merge? Github web gui "Squash and merge" button, or local merge/squash followed by upstream push?

reaperhulk:
Squash and merge in the GitHub UI is how we merge. We have the master branch protected so direct pushes aren't possible without status checks passing.

lmctv:
OK, it was about time!

PR #235, opened 2017-01-10
Fix a small typo in a docstring.

lmctv:
While adding `__hash__` methods in src/nacl/signing.py I stumbled in this happy typo.
Let's fix it.


alex:
🎶

PR #240, opened 2017-01-27
Document ensure and new exceptions

lmctv:


lmctv:
Travis errors stem from the pyenv commit 2657f1049cd . Will be solved if/when #241 gets in

PR #241, opened 2017-01-28
Pin pyenv at the last known-working commit

lmctv:
Commit yyuu/pyenv@2657f1049cd45656918f601096509957d5b74e7c broke virtual environment creation on travis.

Go back to working condition by (temporary) pinning the pyenv version in use to the last working commit.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -3,6 +3,14 @@
 set -e
 set -x
 
+# pin pyenv at the last known good release
+PYENV_COMMIT=99d16707e372143fb35d822c26fe8427719b903c
+
+# the following commit breaks tox virtualenv building
+# on travis-ci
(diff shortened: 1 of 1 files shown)


lmctv:
While yyuu just reverted the travis breaking commit, I think it would be better for the pinning changes to get in, leaving some commented-out instructions
```
# to pin pyenv version, set the PYENV_COMMIT variable
# to the required version commit identifier/tag like in
# PYENV_COMMIT=v1.0.7
```

reaperhulk:
Since we do all this pyenv work already I'm 👍 on merging the ability to pin. Were you planning on updating this PR to reflect the proposed change in your comment?

lmctv:
Will update in a minute.

PR #242, opened 2017-01-28
Complete "We do the py36" PR

lmctv:
@alex forgot adding the creation of python 3.6 pyenv environment in .travis/install.sh while adding py36 to the test matrix in #224 .

I've already rebased the PR onto #241 to show we **really** do *the py36*


reaperhulk:
Ready for a rebase

lmctv:
Rebased

PR #245, opened 2017-01-28
improve coverage

ofek:


reaperhulk:
This touches on a problem with the library that we're discussing in #228

reaperhulk:
The merge of #228 fixes the coverage issue and we're now at 100%

PR #239, opened 2017-01-26
Add some documentation about scrypt password hashing

lmctv:
The trivial merge conflict with #217 can be solved in a couple of minutes.

While writing the documentation, I became aware of a couple of changes I think that could have some merit:
 - If #217 does get merged, I'd like to move the python3.6 compatible scrypt() function to nacl.hashlib,
   which would become a common container for python3.6 compatible implementations
 - I'm unsure I still like the underscore in pw_hash name, and I think I'd prefer to replace the module
   name with pwhash before the next version gets released

PR #247, opened 2017-01-29
[RFC] Move scrypt function to hashlib

lmctv:
As noted in #239, I think nacl.hashlib is a better container both for functions implementing similar APIs to python's stdlib and for "raw" functions, missing libsodium's icing.


reaperhulk:
I don't really have an objection to this since this hasn't appeared in a released version, although if we're going to call it hashlib we need to maintain API compatibility. Are we comfortable with that?

lmctv:
Yes the implemented API is the same, and I don't expect it would change from either side. I kind of like thinking about simply doing (if I don't need tree hashing on the blake2b side)

``` python
try:
    from hashlib import blake2b, scrypt
except ImportError:
    from nacl.hashlib import blake2b, scrypt
```

reaperhulk:
jenkins, retest this please

PR #249, opened 2017-01-29
change a header type for the docs

reaperhulk:


reaperhulk:
We should probably determine if we want cryptography's review rules to apply here, but in this case this PR is docs only and trivial so I'm comfortable merging with passing tests.

PR #250, opened 2017-01-29
add more changelog entries

reaperhulk:


reaperhulk:
(Another trivial docs PR with passing tests)

PR #251, opened 2017-01-29
[RFC] Rename pw_hash to pwhash

lmctv:
I think it's better if we keep ourselves in line with libsodium's namespacing, instead of using different abbreviations, at least when there is no apparent need for such changes.


reaperhulk:
I think there's an instance of pw_hash in the docs that needs to be updated as well?

lmctv:
I was just waiting for this PR to get in to update the docs PR with the new module name

reaperhulk:
Okay, that's fine.

Issue #248, opened 2017-01-29
Document new functionality in /docs


reaperhulk:
https://pynacl.readthedocs.io/en/latest/ is built from the docs dir. Unfortunately we don't have docs for the `nacl.hashlib` module, `scrypt` in general, and the output for [blake2b](https://pynacl.readthedocs.io/en/latest/api/hash/) is not parsing properly. We need to get this fixed up before we can do a 1.1.0 release (docs can be built and viewed locally using `tox -e docs` and then looking at `docs/_build/html/index.html`

reaperhulk:
#239 partially addresses this (scrypt)

lmctv:
I'm beginning to write a nacl.hashlib API doc.

lmctv:
I think most of the new stuff is documented now

reaperhulk:
I think I agree. Anything additional missing we'll just open PRs directly for.

PR #252, opened 2017-01-29
Document hashlib's API

lmctv:
Add some docs about scrypt and blake2b as implemented in nacl.hashlib as requested in #248

### DIFF
--- a/docs/api/hashlib.rst
+++ b/docs/api/hashlib.rst
@@ -1,19 +1,18 @@
 nacl.hashlib
 ============
 
-The :py:mod:`nacl.hashlib` module exposes directly usable implementations
-of raw constructs which libsodium exposes with simplified APIs, like the
-ones in :py:mod:`nacl.hash` and in :py:mod:`nacl.pw_hash`.
-
-When such constructs are exposed in some cpython version standard library,
-we will do our best to keep API compatibility with such implementations.
-
-
-
 .. currentmodule:: nacl.hashlib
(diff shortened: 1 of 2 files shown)


reaperhulk:
I think there's an instance of `pw_hash` in the docs that needs to be updated as well?

docs/api/hashlib.rst
```diff
@@ -0,0 +1,83 @@
+nacl.hashlib
+============
+
+The :py:mod:`nacl.hashlib` module exposes directly usable implementations
+of raw constructs which libsodium exposes with simplified APIs, like the
+ones in :py:mod:`nacl.hash` and in :py:mod:`nacl.pw_hash`.
```
reaperhulk: This needs to be pwhash now

docs/api/hashlib.rst
```diff
@@ -0,0 +1,83 @@
...
+
+When such constructs are exposed in some cpython version standard library,
```
reaperhulk: I don't think this paragraph is what we want to say. The current implementations are compatible with the standard library is all we really need to say. The stdlib is extraordinarily unlikely to break backwards compatibility

lmctv:
@reaperhulk  I think this PR is ready for another review cycle. Since I'd grown unhappy with discussing scrypt before blake2b, I swapped them in my last commit; if the original order seems better... I't just a single  commit to trim before merging.

PR #254, opened 2017-01-30
Update public and secret key encryption docs with optional nonces

felipedau:
Sorry for taking such a long time to work on this.

I was going to create new examples when using optional nonces as @reaperhulk suggested on #210, but I decided to use the same ones because they would not be too different. What do you think?

The other questions I have are:

- Should we have one example using a counter or some other "kind" of nonce? Currently the examples show how to use a random nonce manually and automatically, and I don't think that makes too much sense.
- This PR first explains how to provide a nonce and then how to use a random one. What order should the examples be presented though?
- I noticed there are four paragraphs about nonces in secret key encryption and not much in public key encryption. Should both have the same information about it?

Thanks!

### DIFF
--- a/docs/public.rst
+++ b/docs/public.rst
@@ -65,15 +65,31 @@ This is how the system works:
     #   as just a binary blob of data.
     message = b"Kill all humans"
 
+PyNaCl can automatically generate a random nonce for us, making the encryption
+very simple:
+
+.. code-block:: python
+
+    # Encrypt our message, it will be exactly 40 bytes longer than the
+    #   original message as it stores authentication information and the
+    #   nonce alongside it.
+    encrypted = bob_box.encrypt(message)
+
+However, if we need to use an explicit nonce, it can be passed along with the
+message:
+
+.. code-block:: python
+
     # This is a nonce, it *MUST* only be used once, but it is not considered
     #   secret and can be transmitted or stored alongside the ciphertext. A
     #   good source of nonces are just sequences of 24 random bytes.
     nonce = nacl.utils.random(Box.NONCE_SIZE)
 
-    # Encrypt our message, it will be exactly 40 bytes longer than the
-    #   original message as it stores authentication information and the
-    #   nonce alongside it.
     encrypted = bob_box.encrypt(message, nonce)
+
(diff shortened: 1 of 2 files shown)


reaperhulk:
Logically it makes sense to me to first demonstrate the auto-generation path (with phrasing like "PyNaCl can automatically generate a random nonce on your behalf") and then the explicit nonce path ("However, if you need to pass an explicit nonce here's how").

I don't think it's a big deal to have both examples use random -- users who take the explicit nonce path hopefully know their own use case.

If the information looks the same it'd be nice to have it be consistent, but that should be a separate PR.

felipedau:
> Logically it makes sense to me to first demonstrate the auto-generation path (with phrasing like "PyNaCl can automatically generate a random nonce on your behalf") and then the explicit nonce path ("However, if you need to pass an explicit nonce here's how").
> 
> I don't think it's a big deal to have both examples use random -- users who take the explicit nonce path hopefully know their own use case.

Thanks @reaperhulk, I amended the commit. I think it looks better with your suggestion :)

> If the information looks the same it'd be nice to have it be consistent, but that should be a separate PR.

Alright!

I also noticed the `Reference` sections should be updated as well. Are those generated automatically?

reaperhulk:
@felipedau if you mean the sections below that, no they're not auto-generated. If you could submit a PR to update those that'd be great!

felipedau:
> @felipedau if you mean the sections below that, no they're not auto-generated. If you could submit a PR to update those that'd be great!

Yeap, that's what I meant. Alright! Do you think I can use this same branch for that (as it is still updating the docs for optional nonces) and make a new PR? I honestly do not remember if I ever reused a merged PR branch and even if that is correct.

Thanks @reaperhulk!

reaperhulk:
Since we squash on merge the history will be divergent so I don't recommend it. New branch from updated master.

felipedau:
On Fri, Feb 03, 2017 at 07:49:26AM -0800, Paul Kehrer wrote:
> Since we squash on merge the history will be divergent so I don't recommend it. New branch from updated master.

Okay, sounds good.

Thanks!

PR #256, opened 2017-02-03
Decode both smessage and signature in VerifyKey.verify()

blag:
This PR makes it easier to verify messages by properly combining encoded messages and signatures.

Basically it fixes `VerifyKey.verify()` to decode both the `smessage` and `signature` before appending them together. This PR makes it possible to do this:

```python
smsg = sender_signing_key.sign(msg, encoder=Base64Encoder)

verified_msg_1 = sender_verify_key.verify(smsg, encoder=Base64Encoder)  # The sig+msg as smsg

msg = smsg.message  # b64-encoded
sig = smsg.signature  # b64-encoded

verified_msg_2 = sender_verify_key.verify(msg, sig, encoder=Base64Encoder)  # Split msg and sig

assert verified_msg_1 == verified_msg_2
```

It also adds two tests that make sure both ways work and return the same result.

### DIFF
--- a/src/nacl/signing.py
+++ b/src/nacl/signing.py
@@ -106,10 +106,10 @@ class VerifyKey(encoding.Encodable, StringFixer, object):
         if signature is not None:
             # If we were given the message and signature separately, combine
             #   them.
-            smessage = signature + smessage
-
-        # Decode the signed message
-        smessage = encoder.decode(smessage)
+            smessage = encoder.decode(signature) + encoder.decode(smessage)
+        else:
+            # Decode the signed message
+            smessage = encoder.decode(smessage)
 
         return nacl.bindings.crypto_sign_open(smessage, self._key)
 
--- a/tests/test_signing.py
+++ b/tests/test_signing.py
@@ -23,7 +23,7 @@ import pytest
 from utils import assert_equal, assert_not_equal
 
 from nacl.bindings import crypto_sign_PUBLICKEYBYTES, crypto_sign_SEEDBYTES
-from nacl.encoding import HexEncoder
+from nacl.encoding import Base64Encoder, HexEncoder
 from nacl.exceptions import BadSignatureError
 from nacl.signing import SignedMessage, SigningKey, VerifyKey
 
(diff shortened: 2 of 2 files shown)


blag:
Can I get some feedback on this?

reaperhulk:
cc @dstufft since he did the original encoder/decoder design, but based on the docstring it does appear the encoder should do this.

dstufft:
I feel like I had a reason for this at one point, but I don't actually remember what it was or if it was really important. I don't have a problem with it.

blag:
@dstufft Thanks for the review!

@reaperhulk Let me know if there's anything else you need me to do to get this merged.

lmctv:
@blag I'm unsure we really want to decode a detached signature on input, since it's very simple to think about use cases where a detached signature is encoded in a different way than the input message e.g. the case of a hex-encoded signature for a binary stream. IMHO, it's better to keep the **advanced** use case of detached signatures as it stands, and in line with what we do with detached nonces in SecretBox and Box  encrypt and decrypt methods

blag:
@lmctv I don't understand your argument. If you pass in a raw (binary) smessage and a hex-encoded signature, the current function will still not work.

Here's the current function (without comments):

```python
def verify(self, smessage, signature=None, encoder=encoding.RawEncoder):
    if signature is not None:
        smessage = signature + smessage

    # Decode the signed message
    smessage = encoder.decode(smessage)

    return nacl.bindings.crypto_sign_open(smessage, self._key)
```

what happens is that the hex-encoded signature is concatenated with the raw binary smessage, resulting in an `smessage` that is not correct hex-encoded data or binary data!

The current code currently assumes that both the encoded smessage and signature data can be correctly combined by concatenating them. That assumption doesn't hold true for any encoded smessages.

**The current code is absolutely broken for encoded smessages and/or encoded signatures.**

Furthermore, the functions you point out aren't even consistent among themselves.

The `SecretBox.encrypt` function absolutely does encode the nonce and the ciphertext with the same encoder ([source](https://github.com/pyca/pynacl/blob/master/src/nacl/secret.py#L96)). This PR is consistent with that.

The `SecretBox.decrypt` function assumes that signatures are always raw binary data ([source](https://github.com/pyca/pynacl/blob/master/src/nacl/secret.py#L125)).

Even if you insist on keeping this function consistent with that, this `verify` function is _still_ broken!

In that case, this function should be:

```python
def verify(self, smessage, signature=None, encoder=encoding.RawEncoder):
    if signature is not None:
        smessage = signature + smessage_encoder.decode(smessage)
    else:
        # Decode the signed message
        smessage = smessage_encoder.decode(smessage)

    return nacl.bindings.crypto_sign_open(smessage, self._key)
```

@reaperhulk Please let me know what you would like me to do to fix this function.

lmctv:
@blag please accept my apologies for having been too terse in my comment! IMHO, the suggested change at the end of https://github.com/pyca/pynacl/pull/256#issuecomment-453741138
is a correct bug fix, and stays in line with my suggestion about keeping in line with what happens for Box and SecretBox. As such, I'm inclined to promptly accept a PR fixing the missing decode of the input message in case of detached signatures.

lmctv:
Replaced by #504. Closing.

PR #257, opened 2017-02-05
Expose siphash24 bindings at the API level

lmctv:


reaperhulk:
LGTM, but could you add a changelog entry in the readme noting that siphash24 is available now?

lmctv:
Changelog entry added. Thanks

PR #259, opened 2017-02-10
Update encryption/decryption docstrings with optional nonces

felipedau:
As discussed in #254, the `Reference` sections in the docs should also be updated with optional nonces information. I noticed that the docstrings for the `decrypt` methods did not mention that the nonces could be omitted, so I updated those as well (I hope both changes belong to this PR). Let me know what you think about the wording. Is it clear?

Thanks!

### DIFF
--- a/docs/public.rst
+++ b/docs/public.rst
@@ -159,8 +159,9 @@ Reference
 
     .. method:: encrypt(plaintext, nonce, encoder)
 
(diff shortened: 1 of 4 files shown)


reaperhulk:
LGTM

felipedau:
On Thu, Feb 09, 2017 at 04:57:27PM -0800, Paul Kehrer wrote:
> LGTM

Thanks!

Issue #233, opened 2017-01-08
Expose Argon2


maqp:
[Libsodium documentation](https://download.libsodium.org/doc/password_hashing/the_argon2i_function.html) states they have included the Password Hashing Competition winner, [Argon2](https://www.cryptolux.org/images/0/0d/Argon2.pdf) and talk about it as "state of the art in the design of memory-hard functions". It would be a valuable addition to this toolkit as it's apparently much better than the current alternatives -- scrypt, bcrypt and PBDKF2-HMAC-SHA256. Please add this to API to allow easy password based encryption of data at rest with PyNaCl.

lmctv:
If nobody beats me at it, I'm willing to add argon2 bindings after #216 and #217 are reviewed and accepted; I was mostly waiting for the import of version 1.0.11 of libsodium, which happened on #225 .

maqp:
@lmctv thank you. While it's not a solution for every version of Python, have you noticed Blake2 (and Keccak/SHA3) was added to Python 3.6s' hashlib?

lmctv:
Yes, I notes both script password hashing and the sha3 competition ones you quoted in #926 .
Still, adding a module to a virtual environment is usually simpler than installing a new version or the interpreter.

itdaniher:
@lmctv currently using `argon2-cffi` _and_ `PyNaCl` - would love to see argon2 exposed!

lmctv:
@itdaniher I'm working on argon2i bindings, but I don't think this should slow down 1.1.0 release.

As an alternative, since both the already integrated scrypt bindings, and the upcoming argon2i ones support the {pwhash}_str and the {pwhash}_verify interfaces using modular crypt() serialization, if you need a serialization format to use on interactive authentication, you could start using ``scrypt..._str`` to serialize the password hashes like in

```python
serialized = nacl.pwhash.scryptsalsa208sha256_str(userpass)
```
and test the serialized format before verification like in

```python
proposed_pass = b'proposed'
if serialized.startswith(b'$7$'):
    res = nacl.pwhash.verify_scryptsalsa208sha256(serialized, proposed_pass)
```
and later replace the serialization with
```python
serialized = nacl.pwhash.argon2i_str(userpass)
```
and add a
```python
elif serialized.startswith(b'$argon2i$'):
    res = nacl.pwhash.verify_argon2i(serialized, proposed_pass)
```
clause to gain argon2i support.

[...]

PR #260, opened 2017-02-11
Add argon2i hashes generated via the command line

lmctv:
This is a prerequisite for an upcoming PR meant to add argon2i bindings and closing #233.

@reaperhulk I don't think the python driver I've used to generate the reference data should be added in, but I'm open to any suggestion.

### DIFF
--- a/tests/data/modular_crypt_argon2i_hashes.json
+++ b/tests/data/modular_crypt_argon2i_hashes.json
@@ -1,0 +1,92 @@
+[
+  {
+    "passwd": "gCb0SBnpxkdLZdWe",
+    "mode": "crypt",
+    "dgst_len": 20,
+    "iters": 4,
+    "salt": "kSmmBSy5uyC4",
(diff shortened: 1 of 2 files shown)


reaperhulk:
Drop it in a comment here so I can confirm the vectors locally.

lmctv:
If you think it's needed, I'll replace  calls ti systemrandom with calls to come deterministic PRNGs and resubmit.

reaperhulk:
Eh, no. If you have the code ready that verifies these vectors using the argon2i bindings in libsodium maybe submit that though? Once those tests pass I'm okay with merging this and then you can rebase.

Would be nice if the argon2 official repo's test vectors were more useful for this scenario...

lmctv:
I hope to have test code running by tomorrow at last..

lmctv:
I've replaced the raw hashes since libsodium forces a fixed length for the salt, and doesn't expose a wrapper for the low-level argon2i construct; while I respectfully disagree with this choice jedisct1/libsodium#488, I think we can only adapt and lower our expectations...

lmctv:
The test vectors got already in via #261

PR #261, opened 2017-02-12
Add pwhash argon2i

lmctv:
As suggested by @reaperhulk, I'm early-posting the argon2i bindings I'm working on, to add the needed testing context to the #260 PR and starting to implement #233.

While I think the implementation is reasonably complete, documentation is completely missing.


lmctv:
I don't understand what's happening in jenkin's windows environments, since all of the symbols I'm using seem to be correctly exported, at least from a libsodium's include perspective.

reaperhulk:
libsodium is precompiled on the windows builder and appears to be out of date. Looks like I need to upgrade it.

lmctv:
Thank you for letting me know. Those Jenkins errors got me worried!

reaperhulk:
Jenkins, retest this please

reaperhulk:
Jenkins, retest this please

lmctv:
@reaperhulk  is jenkin's
```
_sodium.obj : error LNK2019: unresolved external symbol _crypto_pwhash_argon2i_bytes_max
```
caused by a too old libsodium library as happened on february?

lmctv:
@reaperhulk I think this is ready for review now (assuming the windows problem stems from outdated libs)

reaperhulk:
No doubt. I haven't had a chance to upgrade the windows builders to 1.0.12 yet.

reaperhulk:
Jenkins, retest this please

reaperhulk:
Eeeh, jenkins is being a pain but hypothetically libsodium is now upgraded. Rebase/push a commit and we can confirm that.

reaperhulk:
Looks like there's an issue on 32-bit related to maximum memory limit and I need to do one fix for 64-bit python 2.7 on my end.

lmctv:
Strange failures:
1. jenkins  py27 {w32,w64}:
  ```
  c:\libsodium-1.0.12-msvc\include\sodium\crypto_hash_sha512.h(12) :
  fatal error C1083: Cannot open include file: 'stdint.h': No such file or directory
   ```
2. jenkins  py3{3,4,5} w32:
  ```
  nacl.exceptions.ValueError: memlimit must be at most 32768 bytes
  ```
Must see what happens with crypto_pwhash_argon2i_MEMLIMIT_MAX on a 32bit linux system.

reaperhulk:
the 64-bit error should be resolved. I had a missing header.

lmctv:
@reaperhulk  Getting better... A least, the workaround for crypto_pwhash_argon2i_memlimit_max() returning 32768 on 32 bit windows lets the tests pass for all the supported python3 version.

On the other hand, python 2.7 is still getting blocked by
```
c:\libsodium-1.0.12-msvc\include\sodium\crypto_hash_sha512.h(12) :
   fatal error C1083: Cannot open include file: 'stdint.h': No such file or directory
```

I'd like to understand why, on win32, the line
```
#define crypto_pwhash_argon2i_MEMLIMIT_MAX ((SIZE_MAX >= 1ULL << 48) ? 4398046510080U : (SIZE_MAX >= 1ULL << 32) ? 2147483648U : 32768U)
```
seemingly results in  `crypto_pwhash_argon2i_MEMLIMIT_MAX` being defined as `32768`, but I have no idea for now. Would you know the exact version of the prebuilt libraries the windows 32 builders are using?

lmctv:
@reaperhulk  I verified the issue with crypto_pwhash_argon2i_memlimit_max returning a very small value is present in both windows and linux 32 bit systems, and opened jedisct1/libsodium#522 upstream.
As for python2.7 on 32 bit windows, it seems the header you added to solve the 64 bt case is needed there too. Thanks.

reaperhulk:
Fixed, sorry!

lmctv:
@reaperhulk thank you for fixing jenkins environment! I think PR is ready for review now.

lmctv:
@reaperhulk I think this is ready for review now.

lmctv:
@reaperhulk do you think you could find some cycles to spare on this and SealedBox?

maqp:
What kind of review is the pull request awaiting: Feedback on API? Some kind of code audit?

lmctv:
Any kind of review would surely help both me, if you find missing or wrong stuff, and the committers in reviewing and eventually approving the PR.

maqp:
Thoughts on key derivation API

* `opslimit` and `memlimit` (and their default values) should probably be explained in the [documentation](https://github.com/lmctv/pynacl/blob/add-pwhash-argon2i/docs/password_hashing.rst#key-derivation)
    
* Argument for adjusting parallelism seems to be missing from the API
        
* I'd love to have all different parameters available in the API so that the [test vectors](https://tools.ietf.org/html/draft-irtf-cfrg-argon2-02#section-6.2) could be replicated easily. (This means parameters for `secret` and `associated data` would also need to be added (preferably between `memlimit` and `encoder`). I understand duplicate tests are not very useful, but it could help users to check that the library is being used correctly.

I found no other major issues in usability. My expertise in reviewing the bindings and implementation are non-existent and IANAC, so proper review is up to others. That being said, thank you for your effort!

lmctv:
You are right about the need to at least expose the named opslimit and memlimit constants; as for the limited API, we have to live within the limits of what libsodium itself exposed; in this case, the simplified argon2i API documented at
https://download.libsodium.org/doc/password_hashing/the_argon2i_function.html

 While preparing this PR, I've even tried to open an issue suggesting to also export the low-level full phc 
API, jedisct1 /libsodium#488 but the issue has been rejected. 

Thank you for your time.

src/nacl/bindings/crypto_pwhash.py
```diff
@@ -223,3 +265,143 @@ def crypto_pwhash_scryptsalsa208sha256_str_verify(passwd_hash, passwd):
...
     return True
+
+
+def crypto_pwhash_argon2i(outlen, passwd, salt, opslimit, memlimit, alg):
+    """
+    Derive a raw cryptographic key using the ``passwd`` and the ``salt``
+    given as input.
+
+    :param outlen: the length of the derived key
+    :type outlen: output key length
```
reaperhulk: This should say int

tests/data/modular_crypt_argon2i_hashes.json
```diff
@@ -0,0 +1,92 @@
+[
```
reaperhulk: I may have asked this before, but remind me where these test vectors are from?

reaperhulk:
Oops, I think #260 is where these test vectors come from actually? I assume you intend to rebase this PR after #260 merges still, is that correct?

lmctv:
I posted two distict PRs to keep myself honest with the trst vectors; but I think this could as well get in all at once.

I've generated he vectors scripting the command line interface from mainline argon2, since the RFC vectors used different sized salts, which are not supported by libsodium; if you think it could be useful, I could add the generator script.

lmctv:
@reaperhulk I just remembered I've already put the driver script online at https://gist.github.com/lmctv/b035a96ca7c867464ab62482dc81555c

To use the script, and generate a set of test vectors, you need a executable 
copy of the argon2 cli utility you get from P-H-C/phc-winner-argon2 .

I'm unsure about what is the right place for the gist script in pynacl sources; any suggestions?

src/nacl/bindings/crypto_pwhash.py
lmctv: I'm away from my workstation for a couple of days; will add the needed change in a few days.

lmctv:
@reaperhulk what about putting the vector generation driver inside the docs?

reaperhulk:
We do that with cryptography, although we don't currently have a place for it in the docs here. If you want to add a new section in the style we use on the other project that would be good.

lmctv:
Just took a look at the general structure of the vectors sectoin of cryptography docs; it seems at least a crude version for pynacl is within reach of my writing skills...

lmctv:
@reaperhulk  would you mind taking a look at the present status of #287 ? I think that would be the right place for adding a "how this project is tested" heading and then describe where the test vectors come from, and the instructions/code needed to generate those vectors we didn't find a canonical source for.

lmctv:
@reaperhulk just a little note on merge strategy: since the single-commit in #260 was not altered in this PR, even if merging separately there should be non need to rebasing #261 after merging #260.

lmctv:
@reaperhulk P.S. for the separate merge to work, I think #260 must be merged as-is and not rebased on top of current master...

docs/password_hashing.rst
```diff
@@ -111,23 +160,71 @@ the decryption would fail and an exception would be raised::
...
+    ...                               guessed_pw, salt,
+    ...                               opslimit=ops, memlimit=mem)
     >>> Eves_box = secret.SecretBox(Eves_key)
     >>> intercepted = Eves_box.decrypt(encrypted)
     Traceback (most recent call last):
         ...
     nacl.exceptions.CryptoError: Decryption failed. Ciphertext failed ...
 
+
+Contrary to the hashed password storage case, where a serialization
```
reaperhulk: Small rephrasing:

Contrary to the hashed password storage case where a serialization format is well-defined, in the raw key derivation case the library user must take care to store (and retrieve) both a reference to the kdf used to derive the secret key and all the derivation parameters. These parameters are needed to later generate the same secret key from the password.

lmctv:
@reaperhulk thank you very much for your continued help and support.
@maqp thank you very much for your helpful pre-review.

PR #263, opened 2017-02-20
Add some notes about modular crypt() format usage

lmctv:
While not strictly needed, I think some notes like the ones I'm proposing could save some time to the library users.


blag:
Crypto is hard enough as it is - adding documentation suggesting good ways to use this library is an excellent idea!

lmctv:
Closing, since this was just meant as a stop-gap for 1.1 . I'll add equivalent notes to #261 .

PR #253, opened 2017-01-29
[RFC] Replace standard library exceptions with nacl defined ones

lmctv:

PR #264, opened 2017-02-20
Correct exceptions' method resolution order

lmctv:
by swapping the mixed-in nacl.exception.CryptoError and the standard library parent exception in class declarations.

With this change the exceptions will "prefer" using the methods inherited from the corresponding standard library one, instead of those inherited from Exception via CryptoError.

This is not strictly needed, since we don't catch and re-raise OSError, but it feels a bit more correct.

### DIFF
--- a/src/nacl/exceptions.py
+++ b/src/nacl/exceptions.py
@@ -27,19 +27,19 @@ class BadSignatureError(CryptoError):
     """
 
 
-class RuntimeError(CryptoError, RuntimeError):
+class RuntimeError(RuntimeError, CryptoError):
     pass
 
 
-class AssertionError(CryptoError, AssertionError):
+class AssertionError(AssertionError, CryptoError):
     pass
 
 
-class TypeError(CryptoError, TypeError):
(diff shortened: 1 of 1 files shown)


blag:
Unless there's a definite and clear reason to do this, I'd rather keep it the way it is. I don't want my crypto software being changed because it "feels" more correct.

lmctv:
@blag I respectfully suggest you take a look at ``git blame src/nacl/exceptions.py``, to see all of those exceptions were introduced in commits 718652a, d5e0e34 and 35792a0, respectively originated from my PRs #223 "Asserts replacer", #232 "Use constant-time comparison..." and #204 "Add crypto pwhash scryptsalsa208sha256" .

Since you feel so strongly for nacl exceptions, I'd be grateful if you could take a look to #253 .

Thank you very much.

blag:
@lmctv Sorry if I offended you. I didn't see any good rationale for changing this, but I didn't realize you're the original author of them. Now that I know that, I trust you to make the right call.

Thank you for all of your work.

PR #266, opened 2017-03-13
Release automation

### LINKED ISSUE #221: Write some release automation
We should lift some of our release automation from cryptography and bcrypt to make releases easier here.

reaperhulk:
Fixes #221

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1,0 +1,131 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import absolute_import, division, print_function
+
+import getpass
+import io
+import os
+import time
+
+from clint.textui.progress import Bar as ProgressBar
+
+import invoke
+
+import requests
+
+
+JENKINS_URL = "https://jenkins.cryptography.io/job/pynacl-wheel-builder"
+
+
+def wait_for_build_completed(session):
+    # Wait 20 seconds before actually checking if the build is complete, to
+    # ensure that it had time to really start.
+    time.sleep(20)
+    while True:
+        response = session.get(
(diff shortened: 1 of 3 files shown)


alex:
I observe that this is basically a copy-paste of the `pyca/cryptography` automatino.

reaperhulk:
Mostly. I fixed a few small things though so I will be sending a PR to improve cryptography's automation later today based on what I did here.

tasks.py
```diff
@@ -0,0 +1,131 @@
...
+
+    invoke.run("twine upload -s dist/PyNaCl-{0}* ".format(version))
+
+    session = requests.Session()
+
+    # This tells the CDN to delete the cached response for the URL. We do this
+    # so that the Jenkins builders will see the new sdist immediately when they
+    # go to build the wheels.
+    response = session.request(
+        "PURGE", "https://pypi.python.org/simple/pynacl/"
```
dstufft: This shouldn't be required if you use Warehouse fwiw. The purging there has fancy things like retries! However it doesn't really hurt to keep it either.

tasks.py
```diff
@@ -0,0 +1,131 @@
...
+        "{0}/build".format(JENKINS_URL),
+        params={
+            "cause": "Building wheels for {0}".format(version),
+            "token": token
+        }
+    )
+    response.raise_for_status()
+    wait_for_build_completed(session)
+    paths = download_artifacts(session)
+    invoke.run("twine upload {0}".format(" ".join(paths)))
```
dstufft: Do we purposely only sign sdists?

reaperhulk: That's how we've done it on cryptography. I don't have any objection to signing more, but I also don't want to type my GPG key password in 20 times :)

dstufft: Fine with me, just noticed it.

PR #274, opened 2017-03-15
Break import cycle (#273)

reaperhulk:
* Move ``ensure`` definition from utils to exceptions

* Use correct import for ``ensure``

* Test relocated ``ensure`` asserts replacer

* Move ``ensure`` tests to a file of their own.


reaperhulk:
This is a backport of an identical merged commit (#273)

PR #279, opened 2017-03-26
Add extended output length `siphashx24` variant of siphash24

lmctv:


lmctv:
The PR is a straightforward extension of the work done for siphash24. With jenkins builds fixed, I think this is ready for review

docs/api/hash.rst
```diff
@@ -60,3 +60,19 @@ nacl.hash
     :param encoder: A class that is able to encode the hashed message.
     :return: The hashed message.
     :rtype: bytes(:py:data:`.SIPHASH_BYTES`) long bytes sequence
+
+
+.. function:: siphashx24(message, key=b'', encoder=nacl.encoding.HexEncoder)
```
reaperhulk: Are we doing `.. versionadded::` in pynacl docs?

lmctv: You're right; I think we should.

docs/api/hash.rst
```diff
@@ -60,3 +60,21 @@ nacl.hash
...
+    :type message: bytes
+    :param key: the message authentication key to be used
+                It must be a :py:data:`.SIPHASHX_KEYBYTES` long
+                bytes sequence
+    :type key: bytes(:py:data:`.SIPHASHX_KEYBYTES`)
+    :param encoder: A class that is able to encode the hashed message.
+    :return: The hashed message.
+    :rtype: bytes(:py:data:`.SIPHASHX_BYTES`) long bytes sequence
+
+    .. versionadded:: 1.2
```
reaperhulk: I'd prefer this to live directly below the function declaration so the docs show the version added information right next to it. Otherwise this LGTM

Issue #281, opened 2017-03-28
undefined symbol: crypto_generichash_blake2b_salt_personal


Josephvb10:
After installing PyNaCl with pip3.5 on Centos 6 and trying to import nacl.secret from a Python script, I get the following error:

```
Traceback (most recent call last):
  File "test.py", line 1, in <module>
    import nacl.secret
  File "/usr/lib64/python3.5/site-packages/nacl/secret.py", line 17, in <module>
    import nacl.bindings
  File "/usr/lib64/python3.5/site-packages/nacl/bindings/__init__.py", line 17, in <module>
    from nacl.bindings.crypto_box import (
  File "/usr/lib64/python3.5/site-packages/nacl/bindings/crypto_box.py", line 18, in <module>
    from nacl._sodium import ffi, lib
ImportError: /usr/lib64/python3.5/site-packages/nacl/_sodium.cpython-35m-x86_64-linux-gnu.so: undefined symbol: crypto_generichash_blake2b_salt_personal
```

lmctv:
Would you mind trying to execute the
```
$ ldd /usr/lib64/python3.5/site-packages/nacl/_sodium.cpython-35m-x86_64-linux-gnu.so
```
command on that system, to try and see which version of the sodium library is getting dynamically linked to  pynacl bindings?

In as much as I know, recent versions of pynacl should statically link the vendored version of libsodium at build time, unless you specifically set the environment variable `SODIUM_INSTALL=system` while building.

Josephvb10:
This is what I'm getting:

# ldd /usr/lib64/python3.5/site-packages/nacl/_sodium.cpython-35m-x86_64-linux-gnu.so
        linux-vdso.so.1 =>  (0x000069d830b69000)
        libsodium.so.4 => /usr/lib64/libsodium.so.4 (0x000069d8306d0000)
        libpython3.5m.so.1.0 => /usr/lib64/libpython3.5m.so.1.0 (0x000069d830200000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x000069d82ffe0000)
        libc.so.6 => /lib64/libc.so.6 (0x000069d82fc48000)
        libm.so.6 => /lib64/libm.so.6 (0x000069d82f9c0000)
        librt.so.1 => /lib64/librt.so.1 (0x000069d82f7b8000)
        libdl.so.2 => /lib64/libdl.so.2 (0x000069d82f5b0000)
[...]

Issue #179, opened 2016-03-03
SODIUM_INSTALL=bundled still links with preinstalled libsodium.so


yurivict:
Observing this on FreeBSD.

yurivict:
Fixed in 1.1.2.

gohonsen:
@yurivict  It seems not effect for centos 6.5 to me;  
the system libsodium version was 0.4.5, the ImportError is " /usr/lib64/python2.6/site-packages/PyNaCl-1.2.0.dev1-py2.6-linux-x86_64.egg/nacl/_sodium.so: undefined symbol: crypto_generichash_blake2b_salt_personal ", the ldd info like this:
`ldd  /usr/lib64/python2.6/site-packages/PyNaCl-1.2.0.dev1-py2.6-linux-x86_64.egg/nacl/_sodium.so
linux-vdso.so.1 =>  (0x00007ffc33f98000)
libsodium.so.4 => /usr/lib64/libsodium.so.4 (0x00007fdfa7c79000)
libpython2.6.so.1.0 => /usr/lib64/libpython2.6.so.1.0 (0x00007fdfa78d3000)
libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fdfa76b5000)
libc.so.6 => /lib64/libc.so.6 (0x00007fdfa7321000)
libm.so.6 => /lib64/libm.so.6 (0x00007fdfa709d000)
librt.so.1 => /lib64/librt.so.1 (0x00007fdfa6e94000)
libdl.so.2 => /lib64/libdl.so.2 (0x00007fdfa6c90000)
libutil.so.1 => /lib64/libutil.so.1 (0x00007fdfa6a8d000)
/lib64/ld-linux-x86-64.so.2 (0x00007fdfa80e7000)`

then I installed libsodium-1.0.12 by manually, it was installed at /usr/local/lib, the system libsodium.so  was at /usr/lib64, after rebuild install pynacl, the ImportError still exist,  and the ldd info does't change. At last, I link the /usr/lib64/libsodium.so to new version, the issue was solved. I do the operations as follows:
`mv /usr/lib64/libsodium.so /usr/lib64/libsodium.so_bak
mv /usr/lib64/libsodium.so.4 /usr/lib64/libsodium.so.4_bak
ln -s /usr/local/lib/libsodium.so.18.2.0 /usr/lib64/libsodium.so.18.2.0
[...]

PR #282, opened 2017-03-30
Reorder link-time library search path

lmctv:
putting bundled libsodium lib directory first.

Confirmed by submitters of #179 and #281 to fix the link errors on both freebsd and redhat 6.


lmctv:
@dstufft would you mind taking a look at this small change?

reaperhulk:
LGTM. This path is only used if we're not trying to use the system libs, so our copy should definitely be prioritized.

PR #287, opened 2017-04-08
Move around and split some docs

lmctv:
To look at the end result of this commit: http://lmctvpynacl.readthedocs.io/en/split-out-changes

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -1,0 +1,31 @@
+Installation
+============
+
+Linux
+-----
+
+PyNaCl relies on `libsodium`_, a portable C library. A copy is bundled
+with PyNaCl so to install you can run:
+
+.. code-block:: console
+
+    $ pip install pynacl
+
+If you'd prefer to use one provided by your distribution you can disable
+the bundled copy during install by running:
+
+.. code-block:: console
+
+    $ SODIUM_INSTALL=system pip install pynacl
+
+.. _libsodium: https://github.com/jedisct1/libsodium
+
+Mac OS X & Windows
+------------------
+
+PyNaCl ships as a binary wheel on OS X and Windows so all dependencies
+are included. Make sure you have an up-to-date pip and run:
+
+.. code-block:: console
+
+    $ pip install pynacl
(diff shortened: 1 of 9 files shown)


lmctv:
@dstufft , @reaperhulk I'm unsure if you just didn't look at these proposed changes, or you prefer the status quo.

reaperhulk:
@lmctv I had not looked at these. I'm not sure if we want to move the changelog or not. Ultimately it will become too big, but right now it is nice that it's present in the readme so that it shows up on PyPI.

lmctv:
If you think it's worthwhile, we could concatenate the README.rst and the CHANGELOG.rst in setup.py, to get the changes shown both in pypi  and in the generated docs.

lmctv:
@reaperhulk with my last added commit, the pypi page should continue showing both the changelog and the terse install notes, developers checking out the sources or browsing them on github will find the customary INSTALL and CHANGELOG files and their content will also show in generated documentation.

INSTALL.rst
```diff
@@ -0,0 +1,31 @@
...
+.. code-block:: console
+
+    $ SODIUM_INSTALL=system pip install pynacl
+
+.. _libsodium: https://github.com/jedisct1/libsodium
+
+Mac OS X & Windows
+------------------
+
+PyNaCl ships as a binary wheel on OS X and Windows so all dependencies
```
reaperhulk: We ship a binary wheel for manylinux1 now too 😀  (but we can note that in a subsequent PR)

lmctv:
@reaperhulk regarding my other in-flight PRs, would you prefer

a. rebasing on top of this commit (maybe cleaner, but we'll lose all reviews' context)
b. creating a separate doc PR after respective merge

reaperhulk:
Uhh, hmm. Let's do a separate PR I think. There's a lot of review context on both of those. Are both of them ready for review again?

lmctv:
Yes, I think both are ready for review. Thanks

reaperhulk:
Okay I'll review them in the morning. A bit too tired to do this tonight still. Thanks for all your hard work!

PR #288, opened 2017-04-11
Switch to using setuptools based commands instead of distutils

dstufft:


reaperhulk:
@dstufft remind me why we want to do this again? Presumably because setuptools is already monkeypatching distutils here and we might as well import the real thing?

dstufft:
@reaperhulk you're actually not getting the setuptools variant of those commands at all currently, because you're importing them before you're importing setuptools, so you're getting the distutils commands before setuptools has a chance to monkeypatch (which isn't true via pip, because YOLO).

PR #290, opened 2017-04-21
Don't care about coverage in tests/

blag:
It doesn't make sense to care about the coverage of the code in `tests/`, only in `nacl/`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,7 +2,6 @@
 branch = True
 source =
     nacl
-    tests/
 
 [paths]
 source =


ofek:
I disapprove. It is bad practice not to test your tests.

alex:
No, we want to track the coverage of our tests. Code that's not run there is every bit the liability.

blag:
I don't see the benefit, can you explain that a bit more or point me to something that does?

alex:
test code is code too, anything that leads to it being not run is relevant.

blag:
Huh, interesting. I hadn't thought about that before. Thanks!

PR #295, opened 2017-06-02
switch pynacl to using jenkinsfile as well

reaperhulk:


alex:
All the non-windows branches are technically unused here, dunno if we care or not.

alex:
Also the tests are failing

reaperhulk:
Yep, gotta figure out what's going on with the whole "nothing working" part. 😁

PR #296, opened 2017-06-02
Jenkins wheel builder

reaperhulk:


lmctv:
@reaperhulk does this mean jenkins will build wheels for PRs too?

reaperhulk:
No this script will be invoked as part of the release process. The purpose is just to get it into source control and then the jenkins job will pull it from here. You can see the cryptography job (https://ci.cryptography.io/job/cryptography-support-jobs/job/wheel-builder/) and the associated wheel builder script (https://github.com/pyca/cryptography/blob/master/.jenkins/Jenkinsfile-cryptography-wheel-builder)

lmctv:
OK, just understood there is a implied 'Release automation:' prefix in the PR title...

.jenkins/Jenkinsfile-pynacl-wheel-builder
```diff
@@ -0,0 +1,121 @@
...
+                        # Jenkins logs in as a non-interactive shell, so we don't even have /usr/local/bin in PATH
+                        export PATH="/usr/local/bin:\${PATH}"
+                        export PATH="/Users/jenkins/.pyenv/shims:\${PATH}"
+
+                        printenv
+
+                        virtualenv .venv -p ${pythonPath[version]}
+                        source .venv/bin/activate
+                        pip install -U wheel # upgrade wheel to latest before we use it to build the wheel
+                        pip wheel pynacl --wheel-dir=wheelhouse --no-binary pynacl
```
alex: Should we have a `CFLAGS=-mmacosx-version-min=10.9` or something?

alex:
Besides that comment, this LGTM

.jenkins/Jenkinsfile-pynacl-wheel-builder
reaperhulk: That won't really do anything for PyNaCl. On cryptography we need it to tell the linker what symbols should be weak linked, but since we're not linking against anything like that here it's unnecessary.

Issue #292, opened 2017-05-09
Missing "seed" functions


mitra42:
The "seed" set of functions appear to be missing, the equivalent of box_seed_keypair(xyz) etc. 

This is one of a set of issues that make it hard to do cross-platform crypto between Python and JavaScript (libsodium), both of which have a (different) subset of the bindings. In this case the JS can turn a seed into a PrivateKey/Public key pair, but Python can't, while the Python can create a PrivateKey object from a privatekey, but the JS can't, so there is no simple way that I can see to transfer ownership of objects (possession of the PrivateKey or seed to PrivateKey) from one to the other. (In the previous RSA environment we can export the private key to a file).

lmctv:
@mitra42 @reaperhulk While I don't really understand the reason why different components should share a private-key or its equivalent "seed", adding the low-level binding should be a no-brainer. On the other hand, I'm really reluctant to adding documented API-level bindings unless someone helps me to write some **very harsh** notes explaining seed-sharing is just equivalent to shared-key crypto, and is (almost always) a bad idea.

mitra42:
I appreciate that there are different opinions of how to do crypto. I'll give you the rationale for two cases I'm using this but the meta-problem is that the different bindings  of libsodium implement different subsets making any kind of cross-platform applications really hard because we are limited to the intersection of the sets of ways to use Crypto that you, and the authors of other bindings,  and that makes libsodium a hard choice to build into a framework that has to work in different situations and all our effort goes into working around two different sets of limitations in the bindings, when the underlying functionality is there in the C library (which is what attracted us to libsodium in the first place).

I'm passing private keys around in three cases. 
a) when I want to access things on multiple devices I want one string that can be used to create the private keys. (Ideally the Seed, since it lets me create a Private and a Signing key from a single string). One way I'll build that seed is using the "mnemonic" function in BIP39, (used in BitCoin) but that only works if I can turn it into a single set of bytes that can work on any environment.

b) when I create private resources e.g. a list that will be published, its given a private key for authentication, and a public key for reading, that means I'm managing a lot of private keys, and they live in a resource (a KeyChain) which is itself encrypted with the owner's private key. That means those keys have to be stored in a format that is independent of whether they will be restored on a Python or Javascript (browser) environment. 

[...]

Issue #182, opened 2016-03-30
expose crypto_box_seal() and crypto_box_seal_open()


iambrosie:
Hi guys,

Would it be possible to explose libsodium's crypto_box_seal() and crypto_box_seal_open()?
As discussed here: https://download.libsodium.org/doc/public-key_cryptography/sealed_boxes.html, they allow for some interesting use cases.

Thanks!

ztane:
Not only that, but the problem with using the standard public key boxes is that if the sender key is compromised later on, then the message can also be recovered; and to use ephemeral keys, one has to decide an encoding scheme to ship the public key along.

reaperhulk:
I have no objection to their addition, so if someone wants to step up and contribute a PR that'd be great!

mitra42:
Sorry, i'm not familiar enough with the code (libsodium,  pynacl or the care needed to avoid inadvertent insecurity) to do that

lmctv:
@reaperhulk I'm volunteering, but I'd like to avoid juggling with too many PRs in flight (hint, hint...)

lmctv:
While I'll be going through the cffi route, this could also be implemented just composing the already present cffi bindings for  crypto_box_keypair(), crypto_box() and crypto_box_open(); what about adding an example of such a composition to the docs?

lmctv:
I've just started implementing the SealedBox high-level API; the preliminary branch is named "add-crypto-box-seal-apis" at lmctv/pynacl

I'd be grateful for any preliminary review, before I start adding the needed tests.

lmctv:
@mitra42 while I can understand your frustration, the Box API is just a well-implemented DH key agreement, which imply both parties need the other's public key to be able to generate the shared key.

mitra42:
Well this is another case of the PyNaCl bindings supporting different subsets of libsodium than the Javascript bindings, making it really really hard to write cross-platform applications.  There are good applications for the Sealed box, (just knowing private key of recipient) which is why I presume it was in libsodium. Yes you can do nasty workarounds specific to PyNaCl (e.g. shipping the public key along with the data) but it seems a pity to make specific workarounds when the underlying C code is already there to do this.

lmctv:
@mitra42 while I concur with your desire to avoid creating new half-baked constructs, the sealed_box one being so straightforward, I think a reimplementation in terms of existing APIs wouldn't take more than another dozen lines, and I'm thinking about doing that just for the sake of tests.

[...]

PR #297, opened 2017-06-03
Add SealedBox public key constructs implementation

lmctv:
As requested in #182 

@mitra42 sorry but, as requested by @reaperhulk #292 will have to wait


lmctv:
@reaperhulk the travis failure should be caused by a random pypi connection problem. Would you mind restarting https://travis-ci.org/pyca/pynacl/jobs/239058582 ?

lmctv:
@maqp would you mind taking a look to this PR and add your review/comments?

lmctv:
@reaperhulk  I think both this and #261 are r4r . Would you think you could find some review cycles to spare?

docs/public.rst
```diff
@@ -99,6 +102,52 @@ Finally, the message is decrypted (regardless of how the nonce was generated):
...
 
 
+nacl.public.SealedBox
+~~~~~~~~~~~~~~~~~~~~~
+
+The :class:`~nacl.public.SealedBox` class encrypts messages addressed
+to a specified key-pair by using ephemeral sender's keypairs, which
+will be discarded just after encrypting a single plaintext message.
+
+This kind of construction allows anonymously sending messages, which
```
maqp: Would "without providing any kind of cryptographic proof of sender's authorship" be better than "anonymously"?

docs/public.rst
```diff
@@ -197,3 +246,35 @@ Reference
...
+
+    .. method:: encrypt(plaintext, encoder)
+
+        Encrypt the message using a Box constructed from an ephemeral
+        key-pair and the receiver key, and encrypt the message.
+
+        The public part of the ephemeral key-pair will be enclosed in the
+        returned ciphertext.
+
+        The ephemeral key-pair will be scrubbed before returning the ciphertext,
```
maqp: It's only the private part of ephemeral key-pair that gets scrubbed.

docs/public.rst
```diff
@@ -197,3 +246,35 @@ Reference
...
+
+        :return bytes: The public part of the ephemeral keypair,
+                       followed by the encrypted ciphertext
+
+    .. method:: decrypt(ciphertext, encoder)
+
+        Decrypt the message using a Box constructed from the receiver key
+        and the ephemeral key enclosed in the ciphertext.
+
+        :param bytes plaintext: The plaintext message to encrypt.
```
maqp: This should probably say ciphertext, not plaintext

docs/public.rst
```diff
@@ -197,3 +246,35 @@ Reference
...
+        :param encoder:  A class that is able to decode the ciphertext.
+
+        :return bytes: The encrypted ciphertext
```
maqp: And this should be plaintext, since we're decrypting

src/nacl/public.py
```diff
@@ -247,3 +247,79 @@ def shared_key(self):
...
+            raise exc.TypeError("SealedBox must be created from "
+                                "a PublicKey or a PrivateKey")
+
+    def __bytes__(self):
+        return self._public_key
+
+    def encrypt(self, plaintext, encoder=encoding.RawEncoder):
+        """
+        Encrypts the plaintext message using a random-generated ephemeral
+        keypair and returns the ciphertext encoded with the encoder.
```
maqp: Function also returns the ephemeral public key

maqp:
@lmctv
I added my comments regarding documentation to files. Other parts look good. Also, the API is really easy and intuitive to use.

lmctv:
@maqp thank you very much for your review! I think I've incorporated all of your suggestions.

lmctv:
@reaperhulk would you mind taking a look to this one too, while waiting for my updates to #261 , since each of #297 and #261 can be effortlessly rebased on top of the other one?

src/nacl/public.py
```diff
@@ -247,3 +247,86 @@ def shared_key(self):
...
+        :param encoder: The encoder used to decode the ciphertext.
+        :return bytes: The original plaintext
+        """
+        # Decode our ciphertext
+        ciphertext = encoder.decode(ciphertext)
+
+        plaintext = nacl.bindings.crypto_box_seal_open(
+            ciphertext,
+            self._public_key,
+            self._private_key,
```
reaperhulk: This will raise a cffi exception if the user instantiates `SealedBox` with a public key and then attempts to decrypt. We probably should raise a TypeError if you call decrypt in that case.

tests/test_sealed_box.py
```diff
@@ -0,0 +1,123 @@
...
+import pytest
+
+from nacl.encoding import HexEncoder
+from nacl.public import PrivateKey, PublicKey, SealedBox
+
+
+VECTORS = [
+    # privalice, pubalice, plaintext, encrypted
+    (
+        b"77076d0a7318a57d3c16c17251b26645df4c2f87ebc0992ab177fba51db92c2a",
```
reaperhulk: Source?

src/nacl/public.py
lmctv: There is already a test for this on the low level bindings side; it's the third clause in test_box_seal_wrong_types(), . I understand the need for a direct test at  the API level..

tests/test_sealed_box.py
lmctv: Access to work environment needed... Will answer in a couple days. Thank you very much for the review.

src/nacl/public.py
lmctv: I added the explicit test for the high level API in 74f01c746

tests/test_sealed_box.py
lmctv: I just found I've added a relevant comment in https://gist.github.com/lmctv/d76a2ebefd3227c461ff5450ea487a11 
```
    # libsodium's test don't provide reference data for
    # the sealed box construction.
    # the following has been manually generated
```
In asmuch as I can remember, I've "played the API" on the command line, then verified the python + SecretBox implementation in the gist could correctly decrypt the manually generated data, and kept that data as a reference for cffi implementation.

This is the main reason why I do a round-trip test in test_sealed_box_encryption.

reaperhulk: Hmm, okay. Could we maybe document a way to generate these vectors out of libsodium for the future? Ideally without using pynacl itself :)

lmctv: My plan, if this can get this in now, is to add such example driver to the custom vectors section we discussed for argon2.

reaperhulk: I'm okay with the docs being added later but I do want to be able to locally create a new vector and confirm it works before I'm comfortable merging this. I think I'm okay with this PR outside of that check now, although I'll give it a final pass once I can do that check.

lmctv: Hope I'll be able to find the needed time in the next few days. Thank you very much for your help.

lmctv: I've put at https://gist.github.com/lmctv/4a6de03c614985da68fee462c993751f a C driver for libsodium seal/seal_open APIs.

After compiling with a command line like 
```
cc sealbox_test.c -o sealbox_test \
                   -Isrc/libsodium/src/libsodium/include/ \
                   -L./build/temp.linux-x86_64-3.5/lib/ -lsodium
```
You can use it to generate test lines on standard output by executing
```
./sealbox_test  N [minmsglen[maxmsglen]
```
and to test existing lines in the format <sk>,<pk>,<msg>,<encrypted> from standard input by executing
```
./sealbox_test
```
A pure libsodium round-trip test can be run like
```
 ./sealbox_test  4 16 64| ./sealbox_test
```
In this case, 4 test lines will get generated and tested, with message lengths between 16 and 64 bytes.

lmctv: For the record, I just tried to insert a vector generated by this driver in VECTORS 
```
    (
        b"886e8e517583c633e93415f5591cc7f85131d74edc1f535fcd3b75d60eb81b9c",
        b"3bae7136e69fa4edb7b56759418e68ec0bba7b80669b889d55cfea2dd82b2362",
        b"7a8b67e028f1c565aa92aee5e144675092b8ef",
        (b"a857f31a360c53ce7407ade7c08de6185dfd402b9a37c2d0d2b3e97931c3c61c1540"
         b"b813f96fb898846a8995f614ae89ad5c0205f6217462316334d09fbe22aa2a7b4a")
    )
```
And the tests are still passing.

lmctv:
With the latest commit, moving the test vector into a separate file, the c driver at my github gist https://gist.github.com/lmctv/4a6de03c614985da68fee462c993751f can be used both to generate new test lines, which can be simply appended to the existing `tests/data/sealed_box_ref.txt` file, and testing that reference data.

reaperhulk:
Thanks, I'll look this over as soon as I can!

lmctv:
Just pushed a forced rebase on the next to last commit; this time it seems travis listened.

tests/test_sealed_box.py
```diff
@@ -0,0 +1,144 @@
...
+def _read_crypto_test_vectors(fname, maxels=0):
+    vectors = []
+    path = os.path.join(os.path.dirname(__file__), "data", fname)
+    with open(path, "rb") as fp:
+        for line in fp:
+            line = line.rstrip()
+            if line[0] == b'#'[0] or not line:
+                continue
+            splt = [x for x in line.split(b'\t')]
+            if maxels:
```
lmctv: I don't think there is a sensible way to force covering the else: clause here.

reaperhulk: It looks like this function is never called with `maxels` < 4, so why even have the branch?

lmctv: It's an almost verbatim copy of `_read_crypto_test_vectors` in `test_generichash`; I just thought it would be better to postpone the refactor /unification to a single version into tests/utile.py to a later, single purpose PR.

lmctv:
@reaperhulk  hopefully ready.

src/nacl/public.py
reaperhulk: As maqp notes, this libsodium method returns ciphertext + public key

lmctv: Following @maqp suggestion, I've added the note about the public part of the sender's ephemeral keypair being sent along in the following couple of lines. Nonetheless, I'm pushing an attempted rewording in a few minutes.

lmctv:
While it seems Travis is not listening build request s from this repository, a PR on my form from the same branch succeeded https://travis-ci.org/lmctv/pynacl/builds/246660246

src/nacl/bindings/crypto_box.py
```diff
@@ -189,3 +190,81 @@ def crypto_box_open_afternm(ciphertext, nonce, k):
...
     return ffi.buffer(plaintext, len(padded))[crypto_box_ZEROBYTES:]
+
+
+def crypto_box_seal(message, pk):
+    """
+    Encrypts and returns a message ``message`` using an ephemeral secret key
+    and the public key ``pk``.
+    The ephemeral public key, which is embedded in the sealed box, is also
+    used, in combination with ``pk``, to derive the nonce needed for the
+    underlyng box construct.
```
reaperhulk: Typo

reaperhulk:
(We'll also want to add this to the changelog)

hoang408:
I got an error here:
from nacl.public import PrivateKey, SealedBox
---> ImportError: cannot import name 'SealedBox'
I am using Python 3.6.  Anyone has any idea?

lmctv:
Basic question: are you sure you're running the (still unreleased on ppypi) git version of PyNaCl?

Can you please run
```
import nacl
print(nacl.__version__)
```
to confirm you're running version 1.2.0.dev1 as expected from a self-built from git distribution?

Thank you.

hoang408:
Hi there, this is what it spits out: '1.1.2'
Any idea on how to fix?

lmctv:
Three options: 
1. follow the build instructions at https://pynacl.readthedocs.io/en/latest/install/ 
2. wait a couple of weeks and install the 1.2.0 we hope to get released by then
3. download the box_seal_from_api.py module at https://gist.github.com/lmctv/d76a2ebefd3227c461ff5450ea487a11 and use that SealedBox implementation until 1.2 is released

hoang408:
Can I have an explanation on why that happens anyways?  Google fails to yield any solution.

lmctv:
You are simply trying to import a class which doesn't exist in the version of the module that's available in your python installation.

hoang408:
I can see SealedBox inside nacl\public.py.  Can you explain to me how that is not 'available' in my installation?

lmctv:
I think you are being somewhat misled by the precedence order between components of the python search path. In particular, I expect the filename you'll get from running
```
import os.path
import nacl
print(nacl.__version__)
print(os.path.join(os.path.dirname(nacl.__file__),'public.py'))
```

will point to a file different from the one implementing SealedBox , which must come from a 2.0 Dev zip/tar/git clone.

hoang408:
This  is what I got:
1.1.2
C:\Program Files\Python36\lib\site-packages\nacl\public.py

That is where I have it installed and inspected.  =/

lmctv:
As a matter of facts, please take a look at the contents of `public.py` as of released 1.1 PyNaCl at https://github.com/pyca/pynacl/blob/1.1.2/src/nacl/public.py to confirm SealedBox absence from that version. I'm repeating myself, if you need SealedBox and are unable to install 1.2 dev. you should simply grab the pure-python implementation, and do some conditional import. Have a nice and productive time with PyNaCl.

hoang408:
Pardon the repeated question.  I am new to this, and was following the PDF's example.

lmctv:
Since we don't put pre-release versions on PyPI, I think you should begin exploring the released API, which is documented at http://pynacl.readthedocs.io/en/stable/public/ .
If you really need SealedBox, just grab the gist module, and, instead of importing both Box and SealrdBox, do something like

```
from nacl.public import PrivateKey
try:
    from nacl.public import SealedBox
except ImportError:
    from box_seal_from_api import SealedBox
```

hoang408:
Thanks much.

hoang408:
I encountered this error: module 'nacl.hash' has no attribute 'HASHER'
Please let me know the work around before v.1.2.0 is released.  Thanks.

lmctv:
This is a documentation error, fixed by #352; the line should just read
```
digest = HASHER(msg, encoder=nacl.encoding.HexEncoder)
￼```
to use the just defined HASHER.

Sorry!

hoang408:
It worked.  Thanks.

hoang408:
I got this error: module 'nacl' has no attribute 'utils', by using this line on page 21/44 of the v.1.2.0 PDF:
auth_key = nacl.utils.random(size=64)

May I have the work around?

hoang408:
I encountered this error: module 'nacl.pwhash' has no attribute 'kdf_argon2i' using the code on page 24./44 of the v.1.2.0 PDF.
Please let me know the work around before v.1.2.0 is released. Thanks.

reaperhulk:
@hoang408 the workaround here is to look at the docs for the released version (http://pynacl.readthedocs.io/en/1.1.2/) instead of the master branch. Alternately you can install off master, but we can't provide ongoing support in this fashion.

PR #303, opened 2017-06-13
Add LIBSODIUM_MAKE_ARGS environment variable

### LINKED ISSUE #302: libsodium slow build time
Hi,

I am running into an issue where installing pynacl on a raspberry pi or in a multiarch docker container is very slow.  By default, `make` (https://github.com/pyca/pynacl/blob/master/setup.py#L163) is run in one thread.  Compile times could be reduced substantially (from >10 minutes to 1 minute in testing) by passing `-j4` or higher to `make`.

Would the maintainers entertain the idea of an environment variable such as `LIBSODIUM_MAKE_ARGS` which could be passed like `LIBSODIUM_MAKE_ARGS=-j4 pip install pynacl`?

uSpike:
By allowing a user to pass LIBSODIUM_MAKE_ARGS, we can decrease the compile time on slow platforms by a considerable amount.

In a `multiarch/alpine:armhf-v3.6` docker container:
* `pip install .` takes about 10 minutes
* `LIBSODIUM_MAKE_ARGS=-j4 pip install .` takes just under 4 minutes

Fixes #302

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -159,14 +159,15 @@ class build_clib(_build_clib):
             cwd=build_temp,
         )
 
+        make_args = os.environ.get('LIBSODIUM_MAKE_ARGS', '').split()
         # Build the library
-        subprocess.check_call(["make"], cwd=build_temp)
+        subprocess.check_call(["make"] + make_args, cwd=build_temp)
 
         # Check the build library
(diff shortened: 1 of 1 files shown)


reaperhulk:
coverage numbers don't make sense so merging :) thanks again

PR #304, opened 2017-06-21
use libsodium_make_args to compile faster on travis

reaperhulk:


reaperhulk:
Okay the scope of this crept a bit. It now does both the parallel make args (which frankly doesn't give a huge boost) and uses the container infrastructure on the bundled builds.

reaperhulk:
This didn't gain enough perf to be worth it.

PR #313, opened 2017-06-28
Insert missing versionadded tag

lmctv:
both for bith the high level SealedBox and the low level bindings


reaperhulk:
codecov having trouble, but these aren't going to affect coverage since they're docs only so merging.

PR #314, opened 2017-06-28
Argon2i - add missing changelog entry

lmctv:


reaperhulk:
codecov having trouble, but these aren't going to affect coverage since they're docs only so merging.

PR #316, opened 2017-06-29
Use the refactored read_crypto_test_vectors()

lmctv:
Janitorial stuff, but I love removing duplicated functions.

### DIFF
--- a/tests/test_signing.py
+++ b/tests/test_signing.py
@@ -15,12 +15,10 @@
 from __future__ import absolute_import, division, print_function
 
 import binascii
-import codecs
-import os
 
 import pytest
 
-from utils import assert_equal, assert_not_equal
+from utils import assert_equal, assert_not_equal, read_crypto_test_vectors
 
 from nacl.bindings import crypto_sign_PUBLICKEYBYTES, crypto_sign_SEEDBYTES
 from nacl.encoding import HexEncoder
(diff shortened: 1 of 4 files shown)


tests/test_signing.py
```diff
@@ -34,23 +32,19 @@ def tohex(b):
...
-                    binascii.unhexlify(x[3].encode("ascii"))[:64],
-                ),
-            })
-
-    return answers
+    # hex-encoded fields on each input line: sk||pk, pk, msg, signature||msg
+    # known answer fields: sk, pk, msg, signature, signed
+    DATA = "ed25519"
+    lines = read_crypto_test_vectors(DATA, delimiter=b':')
+    answ = [(
```
reaperhulk: let's just call return here

lmctv: Corrected.

tests/utils.py
```diff
@@ -27,14 +27,14 @@ def assert_not_equal(x, y):
     assert not (x == y)
 
 
-def read_crypto_test_vectors(fname, maxels=0):
+def read_crypto_test_vectors(fname, maxels=0, delimiter=b'\t'):
```
reaperhulk: Sorry one more comment: could we just update the call sites so we don't need a default value for delimiter here?

lmctv: I agree with updating callers, but I'd rather keep delimiter as a keyword argument. Update coming.

lmctv: Since #317 adds another caller, the one getting last must be rebased on the other.

reaperhulk: Let's land this refactor PR with the update and the other one can be rebased.

lmctv: Just pushed the update.

PR #317, opened 2017-06-30
Add "from_seed" APIs for public key encryption

lmctv:
Should fulfill #292 requirements.


src/nacl/public.py
```diff
@@ -72,27 +72,46 @@ class PrivateKey(encoding.Encodable, StringFixer, object):
...
+        # Verify the given seed type and size are correct
+        if not (isinstance(seed, bytes) and len(seed) == cls.SEED_SIZE):
+            raise exc.TypeError(("PrivateKey seed must be a {0} bytes long "
+                                 "binary sequence").format(
+                                                           cls.SEED_SIZE)
+                                )
+        # generate a raw keypair from the given seed
+        raw_pk, raw_sk = nacl.bindings.crypto_box_seed_keypair(seed)
+        # create a PrivateKey instance from the raw keypair
+        prvk = super(cls.__class__, cls).__new__(cls)
```
reaperhulk: Why not just use the constructor? Is it a significant performance hit to do the scalar mult to re-derive the public key again?

lmctv: This is just from a dislike of repeated work... scalarmult is fast, but why do it twice in a row?
If you think readabiliy counts more than that, I'll go back to using the standard constructor.

reaperhulk: I would prefer to use the standard constructor here, yeah.

lmctv: OK, will update in a few minutes

lmctv: Change pushed.

lmctv:
@reaperhulk @maqp Added the missing docstring to `from_seed` method. Now this should be really ready.

maqp:
The docstring looks good! Great work once again!

lmctv:
@reaperhulk I think this is ready.

lmctv:
@reaperhulk  I hope both this and #320 are ready. Could you try and find some time to look at them?

tests/test_bindings.py
```diff
@@ -244,3 +244,28 @@ def test_box_seal_wrong_types():
     with pytest.raises(TypeError):
         c.crypto_box_seal_open(
             None, A_pubkey, A_secretkey)
+
+
+def test_box_seed_keypair_reference():
+
+    from box_secret_from_seed_ref import test_seed, test_pk, test_sk
```
reaperhulk: I don't think these need to be in a separate file, let's just have them be inline in this file along with the comments.

reaperhulk: Oh, having said that I see they're used in both test_bindings and test_box so nevermind.

Do we have any other python files containing vectors like this? Typically we've made these types of vectors text files that we load and parse.

lmctv: I think I've begun preparing this PR before #297 got merged, and I didn't want to add another open-coded version of test.utils.read_crypto_test_vectors, which would have been necessary to add the needed comments about the reference data. Hope this is better now.

lmctv: OK rebased on master after merge of #316

PR #337, opened 2017-09-01
RFC: reenable CI testing for python2.6

lmctv:
Since there are still downstream users stuck with 2.6, I'd propose to keep testing our code under 2.6 for some more time, at least until this doesn't hinder our pace.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ sudo: required
 
 matrix:
     include:
+        - python: 2.6
+          env: TOXENV=py26 SODIUM_INSTALL=bundled CC=gcc
         - python: 2.7
           env: TOXENV=py27 SODIUM_INSTALL=bundled CC=gcc
         - python: 3.3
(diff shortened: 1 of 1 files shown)


reaperhulk:
I'm opposed to resuming support for Python 2.6. It's unsupported by the Python project and users of it are holding back the entire Python ecosystem. Additionally, BigQuery shows that only 2% of downloads for the latest version are coming from it right now, so it's not significant.

For reference, the query:

```sql
SELECT
  REGEXP_EXTRACT(details.python, r"^([^\.]+\.[^\.]+)") as python_version,
  COUNT(*) as download_count,
FROM
  TABLE_DATE_RANGE(
    [the-psf:pypi.downloads],
    DATE_ADD(CURRENT_TIMESTAMP(), -31, "day"),
    DATE_ADD(CURRENT_TIMESTAMP(), -1, "day")
  )
WHERE
  file.project = 'pynacl' and file.version = '1.1.2'
GROUP BY
  python_version,
ORDER BY
  download_count DESC
LIMIT 100
```

lmctv:
OK, sorry.

reaperhulk:
No need to apologize! This is a discussion 😄 If others disagree we can keep support potentially. Don't feel like my word is law here. You are a maintainer as well

PR #339, opened 2017-09-03
RFC: use "bag of constants" classes to group parameters related to the same construct

lmctv:
I've left compatible module-level names for constructs released in 1.1 , which should be deprecated and then removed if the enum idea flies.


alex:
Use of enum does not make sense here, an enum is a type, where it shoudl be valid to pass any of it's values in a given context, that's not what these are, these are just unrelated constants.

-1 from me.

lmctv:
@alex thank you very much for refreshing my memory on the difference between enumerations and attribute bags!
Would you mind taking another look, now that I defined and used a Constant bag-of-attributes class?

lmctv:
Just rebased on the add-argon2id PR and added the needed corrections to the docs.

lmctv:
Useless. Replaced by #357

PR #342, opened 2017-09-19
Use the provided decoder on the nonce as well.

jimux:
HexEncoding works for the ciphertext and keys, but the nonce will throw an error if it isn't raw. This allowed decoding of the nonce same as the ciphertext.


jimux:
Need to fix issues

PR #343, opened 2017-09-24
Upgrade libsodium to 1.0.14

### LINKED ISSUE #318: Upgrade libsodium to 1.0.14
This would expose the `argon2id` partially data-dependent password hashing construct.

lmctv:
Closes #318 

jedisct1/libsodium#600 affects master branch too.

The workaround for LATEST.tar.gz libsodium tarball has been already removed, since upstream fixed the issue.


reaperhulk:
So we don't need the workaround now that the upstream bug is fixed, does this need a rebase?

lmctv:
I added the workaround in a separate commit I've already stripped out. I don't think there is any need to rebased; just need a indipendent check for correct tar expansion.

jedisct1:
Very sorry for the missing file. These LATEST file are automatically built from the `stable` branch.

The `stable` branch should never be broken. This is the first time it happens; I got tricked by a pattern that was in `.gitignore`, so Git didn't display the missing file when I sync'd the branch with the new release.

This has been fixed, and the task that automatically builds the `LATEST` tarballs from the `stable` branch now performs a distcheck test before updating a tarball.

reaperhulk:
@lmctv so was the configure issue also present in https://github.com/jedisct1/libsodium/releases/download/1.0.14/libsodium-1.0.14.tar.gz ? Because we should only be revving to a released version and not `LATEST`.

lmctv:
The problem manifested only while travis tried to run ./configure from `.travis/install.sh`
```
if [[ $SODIUM_INSTALL == 'system' ]]; then
    wget https://download.libsodium.org/libsodium/releases/LATEST.tar.gz
    tar zxvf LATEST.tar.gz
    cd libsodium-*
    ./configure
 [... ]
```
and the work-around consisted in just inserting the needed calls to `aclocal` and
`automake --add-missing` before running `./configure` (as seen in the logs at https://travis-ci.org/pyca/pynacl/jobs/279175112#L1131)

If we really want to change the tarball version downloaded while building on travis, I could add the needed commit here, or we could add a distinct PR at your discrection; still, since @jedisct1 just confirmed that LATEST.tar.gz comes from the stable branch with only the needed fixes, I'd be oriented to just keep things as they are now.

reaperhulk:
Okay, now I understand. Thanks for clarifying 😄 I'm okay with it as is. I'll review this PR tomorrow morning!

jedisct1:
> was the configure issue also present in https://github.com/jedisct1/libsodium/releases/download/1.0.14/libsodium-1.0.14.tar.gz

No.

Releases are made using the `distcheck` target, so 1.0.14 includes the file.

PR #352, opened 2017-10-07
Fix error in example

lmctv:
by calling the just defined function shortcut instead of a non-existent
attribute of ``nacl.hash``

This fixes an error encountered by hoang408.

### DIFF
--- a/docs/hashing.rst
+++ b/docs/hashing.rst
@@ -30,7 +30,7 @@ Message's creator perspective (:py:func:`~nacl.hash.sha256`,
 
     # define a 1024 bytes log message
     msg = 16*b'256 BytesMessage'
-    digest = nacl.hash.HASHER(msg, encoder=nacl.encoding.HexEncoder)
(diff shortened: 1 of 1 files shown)


docs/hashing.rst
```diff
@@ -30,7 +30,7 @@ Message's creator perspective (:py:func:`~nacl.hash.sha256`,
 
     # define a 1024 bytes log message
     msg = 16*b'256 BytesMessage'
-    digest = nacl.hash.HASHER(msg, encoder=nacl.encoding.HexEncoder)
+    digest = HASHER(msg, encoder=nacl.encoding.HexEncoder)
```
reaperhulk: Should we switch these codeblocks to doctest? That would prevent this sort of bug from occurring.

lmctv: I completely agree with the plan switching the examples to doctests, but I'd rather fix the line now, instead of waiting for the doctests change.

reaperhulk: Okay. I'll file an issue to remind us to switch to doctest.

PR #357, opened 2017-10-14
Namespace pwhash mechanisms

lmctv:
Tidying up and simplifying the interfaces of the exposed password hashing mechanisms.


reaperhulk:
Needs a rebase

lmctv:
Rebased.

lmctv:
@reaperhulk waiting for a green-light before going ahead with docs update.

lmctv:
Since this time I'm pretty convinced the changes make sense, I've imported the constants binding work from #358, added a generic password hash verifier and updated the documentation.
@reaperhulk if this flies, I think we could be on the right track for releasing 1.2 soon.

src/nacl/pwhash/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+PWHASH_SIZE = argon2.PWHASH_SIZE
+SALTBYTES = argon2.SALTBYTES
+BYTES_MAX = argon2.BYTES_MAX
+BYTES_MIN = argon2.BYTES_MIN
+
+assert argon2.ALG_ARGON2_DEFAULT == argon2.ALG_ARGON2ID13
+# since version 1.0.15 of libsodium
+
+MEMLIMIT_MAX = argon2id.MEMLIMIT_MAX
```
reaperhulk: I don't think we should expose these as `nacl.pw_hash.MEMLIMIT_MAX`. libsodium has made argon2id the default, but it still uses namespaced functions like `crypto_pwhash_argon2id_memlimit_max` right?

lmctv: These shims follow libsodium's logic, which defines shims in the upper   crypto_pwhash_ namespace to "lift" functions and constants  from the namespace of the crypto_pwhash_${mech}_ choosen as the default one.

An example: ``include/sodium/crypto_pwhash.h`` 
```
#define crypto_pwhash_MEMLIMIT_MIN crypto_pwhash_argon2id_MEMLIMIT_MIN
SODIUM_EXPORT
size_t crypto_pwhash_memlimit_min(void);
```
If you feel strong against this, we could avoid defining the top level shims, and the verify() dispatcher, but they seem useful.

lmctv:
@spang, @maqp how would you like the password hashing API I'm proposing in this PR?
Same question, if they are still interested in PyNaCl and can spare some time, goes to @dstufft and @warner . Thank you very much for any kind of comment.

lmctv:
@alex I'd be flattered if you could spare some time here!

lmctv:
While I haven't still updated the API documentation, I think the updated section in the docs are http://lmctvpynacl.readthedocs.io/en/latest/password_hashing.html could help in evaluating this proposed change. Thank you all for any help and criticism.

src/nacl/pwhash/__init__.py
lmctv: An example of using the pwhash module level constants would be:
```
>>> from nacl import pwhash
>>> from nacl.pwhash import scrypt
>>> pwhash.str(b'mypass', opslimit=pwhash.OPSLIMIT_INTERACTIVE, memlimit=pwhash.MEMLIMIT_INTERACTIVE)
b'$argon2id$v=19$m=65536,t=2,p=1$o8i1RIGIEPdEbskygmZlng$8JaX4WcDZcgzv1IEV/G+NJZafRQS213YiKLk9zCEI5g'
>>> scrypt.str(b'mypass', opslimit=scrypt.OPSLIMIT_INTERACTIVE, memlimit=scrypt.MEMLIMIT_INTERACTIVE)
b'$7$C6..../....qdeSPD.9L.6lTQD5xLbQAzq4AWkVxsVWJUgxju.hs6B$Zh3by6ga7OZiLlVVh0MeKTRCzDV60XWg2sk9NFd.f1/'
>>>
```

In which both the pwhash and the pwhash.scrypy expose the same API.

lmctv:
@reaperhulk  Just added API level documentation; would really appreciate another review pass.

lmctv:
@reaperhulk Now i think this is really ready. Would you mind another review cycle?

lmctv:
@reaperhulk , @alex do you think you could spare some time on a review? IMHO, this is the last hurdle to release 1.2.

reaperhulk:
I am fine with namespacing things but still not a fan of having an implicit default since unless users pin a version of pynacl they'll be broken when it gets switched in the future. @alex do you have an opinion?

lmctv:
As a matter of fact, a change in defaults for the top level str() would just mean we would need to add the correct dispatch in verify(), but the already existing modular crypt hashes would continue working as verifiers for previously hashed passwords.
On the other hand, the raw kdf construction missing an algorithm identifier is the reason why there is no kdf() defined at top level.

PR #358, opened 2017-10-14
Add missing pwhash constants bindings

lmctv:
Some constants exposed from libsodium have been overlooked while implementing the pwhash family of functions.

The low-level implementation doesn't depend on choices about #357.


lmctv:
Already added via #357 . Closing.

PR #359, opened 2017-10-14
Add argon2id to argon2i changelog entry

lmctv:


lmctv:
@maqp would you mind taking a look at http://lmctvpynacl.readthedocs.io/en/namespace-pwhash-mechanisms/password_hashing.html, and add some more helpful comments to the ones you gave at #261 time?

lmctv:
@reaperhulk I think this should get in before 1.2 release

Issue #356, opened 2017-10-08
Switch to using doctest for examples where appropriate


reaperhulk:
This will prevent issues like #352 from occurring in the future.
- [x] docs/hashing.rst
- [x] docs/public.rst
- [x] docs/secret.rst
- [x] docs/signing.rst
- [x] docs/password_hashing.rst

lmctv:
The following rst files contain examples which should become doctests:
- [ ] docs/public.rst
- [ ] docs/secret.rst
- [ ] docs/signing.rst
- [ ] docs/password_hashing.rst

lmctv:
@reaperhulk if you don't mind, I'd add the check list right in the issue opening comment, hopefully getting a progress indicator.

lmctv:
All the code in documentation is tested now! Thanks @reaperhulk

PR #362, opened 2017-10-22
Use sphinx doctests for pwhash docs examples

lmctv:
Based on #357 since I'm working on that now. This would be a first pass at resolving #356

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,8 @@ matrix:
         - python: 3.6
           env: TOXENV=py36 SODIUM_INSTALL=system CC=clang
         - env: TOXENV=pypy SODIUM_INSTALL=system CC=clang
-        - env: TOXENV=docs
+        - python: 3.6
+          env: TOXENV=docs
         - env: TOXENV=meta
 
 install: .travis/install.sh
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@ except ImportError:
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath("../src"))
+# sys.path.insert(0, os.path.abspath("../src"))
 
 import nacl # flake8: noqa
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@ except ImportError:
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-# sys.path.insert(0, os.path.abspath("../src"))
+#sys.path.insert(0, os.path.abspath('.'))
 
 import nacl # flake8: noqa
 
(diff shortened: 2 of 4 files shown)


lmctv:
Just rebased on 6fa80144ead254b207e69de6a83b076985391d3e .

docs/conf.py
```diff
@@ -18,7 +18,7 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath("../src"))
+# sys.path.insert(0, os.path.abspath("../src"))
```
reaperhulk: ?

.travis.yml
```diff
@@ -50,7 +50,8 @@ matrix:
         - python: 3.6
           env: TOXENV=py36 SODIUM_INSTALL=system CC=clang
         - env: TOXENV=pypy SODIUM_INSTALL=system CC=clang
-        - env: TOXENV=docs
+        - python: 3.6
```
reaperhulk: What's going on here?

docs/password_hashing.rst
```diff
@@ -50,7 +50,10 @@ Password storage and verification
 
 All implementations of the modular crypt hasher `str` function
 internally generate a random salt, and return a hash encoded
-in ascii modular crypt format, which can be stored in a shadow-like file::
+in ascii modular crypt format, which can be stored in a shadow-like file
+
+.. doctest::
+    :pyversion: > 3.3
```
reaperhulk: What does this directive do? It's new to me 😄

.travis.yml
lmctv: After removing the explicit ``basepython...`` declaration from tox.ini, we add an explicit requirement for a python3.6 runtime in travis docs build stanza. This is needed to avoid getting spurious errors generated by the representation of 2.7 ``str`` vs 3.x ``bytes``

docs/password_hashing.rst
lmctv: This is a very nice feature in sphinx which, since version 1.6, lets you skip running the tests if the runtime version doesn't match the specified condition

docs/conf.py
lmctv: If ``../src`` is prepended to sys.path, the runtime would try to import the low-level _sodium bindings from a nonexistent file in src/nacl ignoring the one installed in site-packages by pip/setup.py; for a comparison, see:
```
$ ./venv-35/bin/python -c 'import sys; import os; sys.path.insert(0, os.path.abspath("./src"));from nacl._sodium import ffi, lib; print("OK")'
  File "<string>", line 1, in <module>
ImportError: No module named 'nacl._sodium'
```
vs
```
$ ./venv-35/bin/python -c 'from nacl._sodium import ffi, lib; print("OK")'
OK
```

.travis.yml
reaperhulk: 👍

docs/conf.py
reaperhulk: Sorry, I meant that if this is not necessary/desirable we should delete it rather than commenting it out.

In cryptography's config this is just `os.path.abspath(".")` and is set that way so that we can import a custom extension that is in the same dir.

lmctv: I'd prefer restoring the comment to the same content that was generated by sphinx-quickstart.
Let me know if this can make sense.

PR #365, opened 2017-11-01
Ready to use - update version and copyright date for 1.2 release

lmctv:


reaperhulk:
Okay I think we're ready to do the release. This PR needs to be updated to set the date in the changelog and then we can merge and release!

PR #366, opened 2017-11-01
drop support for python 3.3, which has low use

reaperhulk:
Additionally we've dropped it on other pyca projects and I'd like to not have to worry about it in our jenkins/wheel infra.

### DIFF
--- a/src/nacl/pwhash/__init__.py
+++ b/src/nacl/pwhash/__init__.py
@@ -63,7 +63,6 @@ def verify(password_hash, password):
     in the stored hash
     """
     _decoded_hash = password_hash.decode('ascii')
-    # the bytes.beginswith() method isn't supported by python 3.3
     if _decoded_hash.startswith(argon2id.STRPREFIX.decode('ascii')):
         return argon2id.verify(password_hash, password)
     elif _decoded_hash.startswith(argon2i.STRPREFIX.decode('ascii')):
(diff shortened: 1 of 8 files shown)


alex:
README needs to be udpated. `:pyversion: > 3.3` should probably be `>= 3.4` IMO.

Also a comment in `src/nacl/pwhash/__init__.py` needs updating

reaperhulk:
For posterity:

In the last 30 days PyNaCl has had 1267078 downloads, of which less than 700 were Python 3.3. That's 0.05%

src/nacl/pwhash/__init__.py
```diff
@@ -63,7 +63,6 @@ def verify(password_hash, password):
     in the stored hash
     """
     _decoded_hash = password_hash.decode('ascii')
-    # the bytes.beginswith() method isn't supported by python 3.3
```
alex: I think you fixed the wrong thing :-) We can either leave the comment, or switch to stop decoding everything here.

lmctv: @alex is right. I think it would be better to leave the comment alone for now and add a subsequent commit for replacing decoded hash/decoded strprefix in the dispatcher with straight
```
if  password_hash.startswith(argon2id.STRPREFIX):
     . . .
```
etc.

reaperhulk: Oops, didn't mean to remove this 😄

Issue #305, opened 2017-06-25
Add missing information about test vectors


lmctv:
After finding a place for them, add the needed information for:

- [x] SealedBox
      Add C driver for libsodium #297 
- [x] argon2i
      python driver for the `argon2` executable built from P-H-C/phc-winner-argon2 #261 
- [x] siphash24 and siphashx24 #206 , #257 and #279
- [x] scrypt  #204
- [x] blake2b #215

Could be added as follows-up to #287 or right inside that PR.

lmctv:
Done. Closing, with many thanks to @reaperhulk for his help.

PR #372, opened 2017-11-05
Document test vector generation for blake2b

lmctv:
Another step towards #305 fulfillment. Even though it doesn't touch any code, I'd rather keep this PR out of the 1.2.x stream.


lmctv:
@alex , @reaperhulk could you take a look at this?

lmctv:
@alex, do you think you could also take a look at this?

PR #374, opened 2017-11-14
Urgent - Set readthedocs configuration from a .readthedocs.yml file in the repository

lmctv:
This would rectify a glitch with readthedocs rendering of PyNaCl documentation, which misses our doctests, since the installed sphinx version is too old.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,5 @@
+python:
+   version: 3.5
+   pip_install: true
+   extra_requirements:
+       - docs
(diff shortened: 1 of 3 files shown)


lmctv:
@reaperhulk please take a look at the differences between.https://pynacl.readthedocs.io/en/stable/password_hashing/   and http://pynacl.readthedocs.io/en/1.2.0/password_hashing/ with 1.2.1 released, I think we finally tamed RTD. Thank you very much!

Issue #376, opened 2017-11-28
Expose sodium_pad and sodium_unpad


MuxZeroNet:
https://download.libsodium.org/doc/helpers/padding.html
:smile:

lmctv:
Done in #382

PR #382, opened 2017-12-16
Add bindings for libsodium padding API

lmctv:
Fulfills request in #376 .


lmctv:
@reaperhulk I think this is ready for a first review.

PR #383, opened 2017-12-16
More arithmetics

lmctv:
Constant time increment and general addition for long integers - useful to avoid nonce reuse.

### DIFF
--- a/src/nacl/bindings/utils.py
+++ b/src/nacl/bindings/utils.py
@@ -92,3 +92,59 @@ def sodium_unpad(s, blocksize):
     if rc != 0:
         raise exc.CryptoError("Unpadding failure")
     return s[:u_len[0]]
+
+
+def sodium_increment(inp):
+    """
+    Increment the value of a byte-sequence interpreted
+    as a little-endian big integer
+
+    :param inp: input bytes buffer
+    :type inp: bytes
+    :return: integer value of ``inp`` incremented by one
+    :rtype: bytes
+
+    """
+    ensure(isinstance(inp, bytes),
+           raising=exc.TypeError)
+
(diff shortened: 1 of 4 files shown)


src/nacl/bindings/utils.py
```diff
@@ -39,3 +39,59 @@ def sodium_memcmp(inp1, inp2):
...
+
+
+def sodium_increment(inp):
+    """
+    Increment the value of a byte-sequence interpreted
+    as a little-endian big integer
+
+    :param inp: input bytes buffer
+    :type inp: bytes
+    :return: integer value of ``inp`` incremented by one
```
reaperhulk: This documentation doesn't seem right. The integer interpretation has incremented by one, but this function returns a byte string.

src/nacl/bindings/utils.py
```diff
@@ -39,3 +39,59 @@ def sodium_memcmp(inp1, inp2):
...
+    """
+    compute a modular addition (a + b) mod 2^(8*len) in constant time
+    for a given common length of their respective little-endian encoded
+    byte arrays.
+
+    :param a: input bytes buffer
+    :type a: bytes
+    :param b: input bytes buffer
+    :type b: bytes
+    :return: integer value of (``a`` + ``b``) mod 2^(8*len(``a``))
```
reaperhulk: Same

lmctv: OK, will try to think about a better wording of "a byte buffer containing the little-endian byte encoding of a integer whose value is..." if something better doesn't come up with you.

lmctv:
Rebased on 8c2ef40. Should be ready.

PR #386, opened 2017-12-17
Update docstrings in nacl/hash.py

lmctv:
Some more janitorial work on docstrings.

### DIFF
--- a/src/nacl/hash.py
+++ b/src/nacl/hash.py
@@ -11,26 +11,46 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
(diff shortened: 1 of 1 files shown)


src/nacl/hash.py
```diff
@@ -11,67 +11,106 @@
...
+"""
 
 from __future__ import absolute_import, division, print_function
 
 import nacl.bindings
 import nacl.encoding
 
+
 BLAKE2B_BYTES = nacl.bindings.crypto_generichash_BYTES
+"""Default digest size for :func:`blake2b` hash"""
```
reaperhulk: I never use docstrings like this so you'll have to teach me: they go after the line?

lmctv: Yes, they must go after the line they refer to

PR #390, opened 2017-12-22
Replace code blocks with testcode in signing.rst

lmctv:
Part of #356 requirements


lmctv:
@reaperhulk , @alex I think the whole set of PRs from #389 to #392 is ready for review and integration.
Since each PR touches a different file and they can therefore merged in any order, I'd suggest starting with #390, since it also fixes a new link-check error.

PR #393, opened 2018-01-07
test_secret: exercise SecretBox message-corruption exceptions

warner:
Apparently our tests weren't checking that a corrupted SecretBox message was causing `decrypt` to throw a `CryptoError`.


warner:
whoops, forgot how pedantic flake8 was. fixed.

PR #403, opened 2018-02-05
Expand on Ed25519 singing/verification example

kitterma:
This may seem like overkill, but getting confused about this point cost me several hours, so I think it'd be good to be explicit about it in the documentation.


lmctv:
Funny, even after #235 , music is still all over this project!
On a more serious note, adding a note does no harm, if you could correct the typo, this LGTM.

kitterma:
Assuming I found the typo you meant, fixed.

lmctv:
Merged, thank you.

PR #408, opened 2018-02-23
Add bindings to low-level ed25519 arithmetic

Lapin0t:
This PR should be merged after #407 (i can rebase when it is so only the second commit is left to merge here). It adds bindings in `nacl.bindings` to the new functions and symbols (`crypto_scalarmult_ed25519_*` and `crypto_core_ed25519_*`).

### DIFF
--- a/src/bindings/crypto_core.h
+++ b/src/bindings/crypto_core.h
@@ -1,4 +1,4 @@
-/* Copyright 2013 Donald Stufft and individual contributors
+/* Copyright 2017 Donald Stufft and individual contributors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


lmctv:
From a cursory look, I'm inclined to green-light this, as soon as the missing tests for these new bindings will be added. Thanks.

src/bindings/crypto_core.h
```diff
@@ -0,0 +1,23 @@
+/* Copyright 2017 Donald Stufft and individual contributors
```
lmctv: I think you should write 2018 here, since you published the file this year.

src/nacl/bindings/crypto_core.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2013 Donald Stufft and individual contributors
```
lmctv: Also 2018.

lmctv:
@Lapin0t do you think you could find some time to complete this PR with the needed tests?

Lapin0t:
Yeah, i'll try to do that but it'll probably take a couple weeks before i can get to it.

lmctv:
OK, take your time.

lmctv:
Just pinging @Lapin0t in the hope he can find some time to complete the PR

Lapin0t:
@lmctv Yeah it's kinda lame because it would probably take no more than a couple hours to complete. By the end of next week my (student's -- exam rush) duties will end so i should be able to do it then. Not forgetting but sorry for the delay ;)

lmctv:
@Lapin0t I was asking, since the PR seems so nearly finished you really deserve being the one putting the last touches in!

lmctv:
The arithmetic part of the PR contents have been ported to #493 where I also added the needed tests.
from_uniform is unmergeable until someone would find working reference vectors for the elligator implementation.

PR #415, opened 2018-03-20
Added crypto_kx bindings

shawnblakesley:
I needed `crypto_kx` bindings for something I was working on, and wanted to share.

### DIFF
--- a/src/bindings/crypto_kx.h
+++ b/src/bindings/crypto_kx.h
@@ -1,0 +1,38 @@
+/* Copyright 2018 Donald Stufft and individual contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+size_t crypto_kx_publickeybytes(void);
+size_t crypto_kx_secretkeybytes(void);
+size_t crypto_kx_seedbytes(void);
+size_t crypto_kx_sessionkeybytes(void);
+
+int crypto_kx_seed_keypair(unsigned char pk[32],
+                           unsigned char sk[32],
+                           const unsigned char seed[32]);
+
+int crypto_kx_keypair(unsigned char pk[32],
+                      unsigned char sk[32]);
+
+int crypto_kx_client_session_keys(unsigned char rx[32],
+                                  unsigned char tx[32],
+                                  const unsigned char client_pk[32],
+                                  const unsigned char client_sk[32],
+                                  const unsigned char server_pk[32]);
+
+int crypto_kx_server_session_keys(unsigned char rx[32],
+                                  unsigned char tx[32],
+                                  const unsigned char server_pk[32],
+                                  const unsigned char server_sk[32],
+                                  const unsigned char client_pk[32]);
--- a/src/nacl/bindings/__init__.py
+++ b/src/nacl/bindings/__init__.py
@@ -63,13 +63,13 @@ from nacl.bindings.crypto_hash import (
     crypto_hash_sha256_BYTES, crypto_hash_sha512, crypto_hash_sha512_BYTES,
 )
 from nacl.bindings.crypto_kx import (
-    crypto_kx_keypair,
-    crypto_kx_client_session_keys,
-    crypto_kx_server_session_keys,
     crypto_kx_PUBLIC_KEY_BYTES,
     crypto_kx_SECRET_KEY_BYTES,
     crypto_kx_SEED_BYTES,
     crypto_kx_SESSION_KEY_BYTES,
+    crypto_kx_client_session_keys,
+    crypto_kx_keypair,
+    crypto_kx_server_session_keys,
 )
 from nacl.bindings.crypto_pwhash import (
     crypto_pwhash_ALG_ARGON2I13,
--- a/src/nacl/bindings/__init__.py
+++ b/src/nacl/bindings/__init__.py
@@ -69,6 +69,7 @@ from nacl.bindings.crypto_kx import (
     crypto_kx_SESSION_KEY_BYTES,
     crypto_kx_client_session_keys,
     crypto_kx_keypair,
+    crypto_kx_seed_keypair,
     crypto_kx_server_session_keys,
 )
 from nacl.bindings.crypto_pwhash import (
@@ -220,6 +221,7 @@ __all__ = [
     "crypto_generichash_blake2b_state_copy",
 
     "crypto_kx_keypair",
+    "crypto_kx_seed_keypair",
     "crypto_kx_client_session_keys",
     "crypto_kx_server_session_keys",
     "crypto_kx_PUBLIC_KEY_BYTES",
--- a/src/nacl/bindings/__init__.py
+++ b/src/nacl/bindings/__init__.py
@@ -62,6 +62,15 @@ from nacl.bindings.crypto_hash import (
     crypto_hash, crypto_hash_BYTES, crypto_hash_sha256,
     crypto_hash_sha256_BYTES, crypto_hash_sha512, crypto_hash_sha512_BYTES,
 )
+from nacl.bindings.crypto_kx import (
+    crypto_kx_keypair,
+    crypto_kx_client_session_keys,
+    crypto_kx_server_session_keys,
+    crypto_kx_PUBLIC_KEY_BYTES,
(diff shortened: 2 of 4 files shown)


src/bindings/crypto_kx.h
```diff
@@ -0,0 +1,38 @@
...
+size_t crypto_kx_seedbytes(void);
+size_t crypto_kx_sessionkeybytes(void);
+
+int crypto_kx_seed_keypair(unsigned char pk[32],
+                           unsigned char sk[32],
+                           const unsigned char seed[32]);
+
+int crypto_kx_keypair(unsigned char pk[32],
+                      unsigned char sk[32]);
+
```
lmctv: This seems just a renamed x25519 keypair, generator, duplicating nacl.bindings.crypto_box_keypair()

src/bindings/crypto_kx.h
```diff
@@ -0,0 +1,38 @@
...
+size_t crypto_kx_publickeybytes(void);
+size_t crypto_kx_secretkeybytes(void);
+size_t crypto_kx_seedbytes(void);
+size_t crypto_kx_sessionkeybytes(void);
+
+int crypto_kx_seed_keypair(unsigned char pk[32],
+                           unsigned char sk[32],
+                           const unsigned char seed[32]);
+
```
lmctv: OK, since this is a reimplementation of crypto_box_seed_keypair in terms of blake2 instead of sha512.

src/nacl/bindings/__init__.py
```diff
@@ -210,6 +219,14 @@
     "crypto_generichash_blake2b_final",
     "crypto_generichash_blake2b_state_copy",
 
+    "crypto_kx_keypair",
```
lmctv: ditto

src/bindings/crypto_kx.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+size_t crypto_kx_publickeybytes(void);
+size_t crypto_kx_secretkeybytes(void);
```
lmctv: These are just a disguised (crypto_box_PUBLICKEYBYTES, crypto_box_SECRETKEYBYTES) pair.

shawnblakesley:
@lmctv 
While most of these are duplicates of crypto_box, the main usage that I am aiming to add are the two session key functions `crypto_kx_client_session_keys` and `crypto_kx_server_session_keys`. It may make some sense that we could use the box keys here, but I think having the kx key generation functions adds to code clarity when trying to generate those session keys (as jedisct1 mentions in [this PR for libsodium](https://github.com/jedisct1/libsodium/issues/639)).

src/nacl/bindings/crypto_kx.py
```diff
@@ -0,0 +1,150 @@
...
+    secret_key = ffi.new("unsigned char[]", crypto_kx_SECRET_KEY_BYTES)
+    if seed is None:
+        res = lib.crypto_kx_keypair(public_key, secret_key)
+    else:
+        ensure(isinstance(seed, bytes) and
+               len(seed) == crypto_kx_SEED_BYTES,
+               'Seed must be a {0} byte long bytes sequence'.format(
+                   crypto_kx_SEED_BYTES),
+               raising=exc.TypeError)
+        res = lib.crypto_kx_seed_keypair(public_key, secret_key, seed)
```
lmctv: If we want the low-level API to mimic libsodium's one, I'd rather keep separate bindings:
```
def crypto_kx_keypair():
    ". . ."
def crypto_kx_keypair(seed):
    ". . ."
```

src/nacl/bindings/crypto_kx.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def crypto_kx_keypair(seed=None):
+    """
+    Generate a keypair.
+    :param seed: random seed
+    :type seed: bytes
+    :return: (public_key, secret_key)
+    :rtype: (bytes, bytes)
+    """
```
lmctv: I think a comment about the keypair really being a x25519 keypair should be added, either inside the docstring or before the implementation.

tests/test_kx.py
```diff
@@ -0,0 +1,100 @@
...
+        seeded_same = b.crypto_kx_keypair(seed2)
+        assert seeded == seeded_same
+    else:
+        seeded = b.crypto_kx_keypair(seed1)
+        seeded_other = b.crypto_kx_keypair(seed2)
+        assert seeded != seeded_other
+
+
+@pytest.mark.parametrize('execution_number', range(100))
+def test_crypto_kx_session_keys(execution_number):
```
lmctv: Also use `hypothesis` here

tests/test_kx.py
```diff
@@ -0,0 +1,100 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+import pytest
+
+import nacl.bindings as b
+import nacl.exceptions as exc
+
+
+@pytest.mark.parametrize(("seed1", "seed2"), [
```
lmctv: I think it would be better to just generate some random seeds couples via hypothesis' `@given` like we already do in `tests/test_bindings.py`:
```
@given(binary(min_size=32,
              max_size=32),
       binary(min_size=32,
              max_size=32),
       )
@settings(max_examples=20)
def test_crypto_kx_keypair(seed1, seed2):
    ". . ."
```
for the standard case, and a couple of explicit failures cases

src/nacl/bindings/crypto_kx.py
```diff
@@ -0,0 +1,150 @@
...
+                                  client_public_key):
+    """
+    Generate session keys for the server.
+    :param server_public_key:
+    :type server_public_key: bytes
+    :param server_secret_key:
+    :type server_secret_key: bytes
+    :param client_public_key:
+    :type client_public_key: bytes
+    :return: (decryption_key, encryption_key)
```
lmctv: I don't like the renaming of the shared key directions `rx` and `tx` to `decryption` and `encryption`, and i fear it could be misleading for downstream users; better keep with upstream choices `rx_key` and `tx_key`

tests/test_kx.py
```diff
@@ -0,0 +1,100 @@
...
+    server_decryption_key, server_encryption_key = \
+        b.crypto_kx_server_session_keys(s_keys[0], s_keys[1], c_keys[0])
+    client_decryption_key, client_encryption_key = \
+        b.crypto_kx_client_session_keys(c_keys[0], c_keys[1], s_keys[0])
+
+    assert client_decryption_key == server_encryption_key
+    assert server_decryption_key == client_encryption_key
+
+
+def test_crypto_kx_session_keys_wrong_length():
```
lmctv: Not a native english speaker, take this with a grain of salt, but I'd use ...wrong_key_lengths instead.

src/nacl/bindings/crypto_kx.py
```diff
@@ -35,25 +35,41 @@
...
+    ensure(res == 0, "Key generation failed.", raising=exc.CryptoError)
+
+    return (ffi.buffer(public_key, crypto_kx_PUBLIC_KEY_BYTES)[:],
+            ffi.buffer(secret_key, crypto_kx_SECRET_KEY_BYTES)[:])
+
+
+def crypto_kx_seed_keypair(seed):
+    """
+    Generate a keypair with a given seed.
+    This is a duplicate crypto_box_seed_keypair, but
```
lmctv: This comment can be fixed later; while the functionality is duplicated,  the construction is different, since crypto_kx_seed_keypair() uses the blake2b hash primitive instead of sha512.

src/nacl/bindings/crypto_kx.py
```diff
@@ -55,8 +56,9 @@ def crypto_kx_keypair():
 def crypto_kx_seed_keypair(seed):
     """
     Generate a keypair with a given seed.
-    This is a duplicate crypto_box_seed_keypair, but
-    is included for api consistency.
+    This is functionally the same as crypto_box_seed_keypair, however
```
lmctv: Good!

src/nacl/bindings/crypto_kx.py
```diff
@@ -38,8 +38,9 @@
 def crypto_kx_keypair():
     """
     Generate a keypair.
-    This is a duplicate crypto_box_keypair, but
-    is included for api consistency.
+    This is functionally the same as crypto_box_keypair, however
```
lmctv: Sorry for not having been clear enough; the previous version of the docstring was right. The only docstring needing changes was the one for `crypto_kx_seed_keypair`

shawnblakesley: No worries. My bad.

shawnblakesley:
@lmctv looks like there was a build failure on one of the machines, but it's not related to my changes. Any ideas on what to do?

lmctv:
Sometimes travis' outgoing network connections get flaky, and builds randomly fail. I just restarted the failed build, and it correctly ran to completion.

Issue #370, opened 2017-11-02
some of the tests take longer than expected on slow machines -> FAIL


ignatenkobrain:
```python
=================================== FAILURES ===================================
_________________________ test_argon2i_str_and_verify __________________________
    @given(text(alphabet=PASSWD_CHARS, min_size=5, max_size=20),
>          integers(min_value=4,
                    max_value=6),
           integers(min_value=1024 * 1024,
                    max_value=16 * 1024 * 1024)
           )
    @settings(deadline=1500, max_examples=20)
    def test_argon2i_str_and_verify(password, ops, mem):
tests/test_pwhash.py:409: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python2.7/site-packages/hypothesis/executors.py:58: in default_new_style_executor
    return function(data)
/usr/lib/python2.7/site-packages/hypothesis/core.py:136: in run
    return test(*args, **kwargs)
tests/test_pwhash.py:409: in test_argon2i_str_and_verify
    integers(min_value=4,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
args = ('   𧰒 ', 5, 15794431), kwargs = {}, start = 1509610662.526979
result = None, runtime = 1874.4771480560303
    @proxies(test)
    def timed_test(*args, **kwargs):
        self.__test_runtime = None
        start = time.time()
        result = test(*args, **kwargs)
        runtime = (time.time() - start) * 1000
        self.__test_runtime = runtime
        if self.settings.deadline is not_set:
            if (
                not self.__warned_deadline and
                runtime >= 200
            ):
                self.__warned_deadline = True
                note_deprecation((
                    'Test took %.2fms to run. In future the default '
                    'deadline setting will be 200ms, which will '
                    'make this an error. You can set deadline to '
                    'an explicit value of e.g. %d to turn tests '
                    'slower than this into an error, or you can set '
                    'it to None to disable this check entirely.') % (
                        runtime, ceil(runtime / 100) * 100,
                ))
        elif runtime >= self.current_deadline:
>           raise DeadlineExceeded(runtime, self.settings.deadline)
[...]

PR #416, opened 2018-03-21
Allow tests/test_pwhash.py deadlines to be adjusted

kitterma:
For anyone who doesn't care (and thus doesn't set the new environment variable), this change is a no-op.  I've verified the tests all run and pass on both python2.7 and python3.6.

For those that do care (in my case as a Debian package maintainer) we can set the environment variable on a per-architecture basis to allow more time for slower processors.

This should address https://github.com/pyca/pynacl/issues/370 .  Fedora should be able to turn the same knob as needed to address their slow builders.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,9 @@ Changelog
   on fixed-precision big integers represented as little-endian
   byte sequences.
 * Add low-level bindings for the ISO/IEC 7816-4 compatible padding API.
+* Add PYNACL_HYPOTHESIS_DEADLINE environment variable to allow the allowed
+  time for tests in tests/test_pwhash.py to be adjusted for slower processor
+  architectures.  GitHub issue #370
 
 1.2.1 - 2017-12-04
 ------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,9 +12,9 @@ Changelog
   on fixed-precision big integers represented as little-endian
   byte sequences.
 * Add low-level bindings for the ISO/IEC 7816-4 compatible padding API.
-* Add PYNACL_HYPOTHESIS_DEADLINE environment variable to allow the allowed
-  time for tests in tests/test_pwhash.py to be adjusted for slower processor
-  architectures.  GitHub issue #370
+* Set hypothesis deadline to None in tests/test_pwhash.py to avoid
+  incorrect test failures on slower processor architectures.  GitHub 
+  issue #370
 
 1.2.1 - 2017-12-04
 ------------------
--- a/tests/test_pwhash.py
+++ b/tests/test_pwhash.py
@@ -32,11 +32,6 @@ import nacl.encoding
 import nacl.exceptions as exc
 import nacl.pwhash
 
-_deadline = os.getenv('PYNACL_HYPOTHESIS_DEADLINE', 1500)
-# Set the environment variable PYNACL_HYPOTHESIS_DEADLINE to an
-# appropriate value for the architecture under test before running tests.
-# 1,500 milliseconds is the default for fast architectures like amd64.
-
 _all_unicode = u''.join(unichr(i) for i in range(sys.maxunicode))
 PASSWD_CHARS = u''.join(c for c in _all_unicode
                         if (ud.category(c).startswith('L') or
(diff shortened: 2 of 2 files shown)


alex:
I'm not wild about dealing an environment variable in our tests; does this make more sense as an upstream hypothesis feature? We're far from the only folks whose test time will very strongly by CPU speed.

kitterma:
The default (which is why I used os.getenv('PYNACL_HYPOTHESIS_DEADLINE', 1500)) doesn't require any environment variable to be set, so for your testing, it's not needed.  For us (Debian) it's a pretty normal way to vary things by architecture.

I didn't think of another way to solve it without hypothesis (or even worse PyNACL) having to have a list of architectures and values.  In Debian it's not rare at all for new archs to appear.  We currently build PyNaCl for 20 architectures and I'm sure no one else wants to have to deal with any of that.  I could ask the hypothesis people to add an environment variable to allow us to do this, but whether they add it or you add it, it's the same amount of an environment variable in your tests.

Currently we are patching PyNaCl with a single value that's high enough to pass everywhere.  I can keep doing that, but it pretty much defeats the purpose of the tests.  

Would you mind accepting this for now (as PyNaCl is the only place I've run into the problem) and I'll see if we can do something better in the long run in hypothesis?

lmctv:
Since the strong integration between pytest and hypothesis which would be needed to globally set the deadline from the pytest command line is missing at this time, I'm inclined to accept this PR. Any objections from @pyca/pynacl-core?

reaperhulk:
Hypothesis documentation suggests that this setting is going away entirely: http://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.timeout

Seems like we may need to just set it to `hypothesis.unlimited` for future compatibility?

kitterma:
On April 2, 2018 2:56:08 PM UTC, Paul Kehrer <notifications@github.com> wrote:
>Hypothesis documentation suggests that this setting is going away
>entirely:
>http://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.timeout
>
>Seems like we may need to just set it to `hypothesis.unlimited` for
>future compatibility?

That was done in a previous PR.  This is about deadline, which is also problematic on slow architectures.

Scott K

reaperhulk:
Oh, ugh, well one day I'll learn how to read.

What's the consequence of setting deadline=None here? Presumably you're tuning your deadline such that it doesn't fail on slower arches, but it seems like there isn't really much point to a deadline at all for our use case here.

kitterma:
I tested it, and it runs fine.  From a distributor perspective (Debian in my case), I think None is fine.  We have overall build time limits to keep a hung build from blocking things indefinitely and we trust upstream to have dealt with performance regressions.  When I developed this PR, I assumed that the deadline was useful for upstream to trigger on performance regression during development.  If that's not the case, then None should be OK.

kitterma:
None is actually easier for us because then I don't have to worry about setting per-architecture limits, so whether it's better upstream to have a deadline we can over-ride, or just have it set to None and we don't have to care, we can manage.  If None is OK for upstream, I'd prefer that since it's one less thing to worry about.

va1entin:
+1
In NixOS we currently patch the deadline to None and timeout to unlimited as well for pynacl.

reaperhulk:
Okay, I'd prefer `None`. @lmctv do you have an opinion?

lmctv:
I'd be grateful if @ignatenkobrain could confirm whether @reaperhulk's suggestion would work also from a fedora perspective. If so, I'm already approving a simplification commit putting deadline=None in all the `@settings` decorators.

kitterma:
I don't know if @ignatenkobrain is still monitoring this or not, so I checked and the current Fedora RPM spec file just ignores tests results on arm to work around this issue:

https://src.fedoraproject.org/cgit/rpms/python-pynacl.git/tree/python-pynacl.spec

So worst case, they continue to do that.

Scott K

lmctv:
Sorry for my confused wording; my call to @ignatenkobrain was just a wake-up one, since the case put forward by yourself and @va1entin is clear enough. Please, remove the _deadline definition, set deadline=None in all the @settings declarations, and we'll be done.

kitterma:
Will do.  It may tomorrow or Friday before I have time to get to this.

Scott K

kitterma:
Should be good to go now.

lmctv:
@kitterma thank you.

ignatenkobrain:
This should be good for us, thanks!

Issue #406, opened 2018-02-21
ImportError: Undefined symbol "crypto_pwhash_alg_argon2id13"


ex3me0:
**OS:** FreeBSD 12
**Package:** PyNaCl==1.2.1 (installed without SODIUM_INSTALL=system, just pip install pynacl via virtualenv)
**System libsodium:** libsodium-1.0.12
```
Python 2.7.13 (default, Aug 29 2017, 16:07:39)
[GCC 4.2.1 Compatible FreeBSD Clang 5.0.0 (branches/release_50 311606)] on freebsd12
Type "help", "copyright", "credits" or "license" for more information.
>>> from nacl._sodium import ffi, lib
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: /home/ex/venv/lib/python2.7/site-packages/nacl/_sodium.so: Undefined symbol "crypto_pwhash_alg_argon2id13"
```
Seems python using bundled libsodium, but I see error like in case with a system library

ex3me0:
I did some investigation, which proves that _sodium.so requires system version of libsodium.
1. Rename `/usr/local/lib/libsodium.so.18.2.0`
2. Run: `python -c "from nacl._sodium import ffi, lib"`
`ImportError: Shared object "libsodium.so.18" not found, required by "_sodium.so"`

ex3me0:
```
(venv) $ ldd /venv/lib/python2.7/site-packages/nacl/_sodium.so
/venv/lib/python2.7/site-packages/nacl/_sodium.so:
        libthr.so.3 => /lib/libthr.so.3 (0x801622000)
        libsodium.so.18 => /usr/local/lib/libsodium.so.18 (0x80184a000)
        libpython2.7.so.1 => /usr/local/lib/libpython2.7.so.1 (0x801ac1000)
        libc.so.7 => /lib/libc.so.7 (0x800825000)
        libintl.so.8 => /usr/local/lib/libintl.so.8 (0x801ea5000)
        libutil.so.9 => /lib/libutil.so.9 (0x8020b0000)
        libm.so.5 => /lib/libm.so.5 (0x8022c4000)
```

reaperhulk:
Yeah, in the past this was a linker search path ordering issue (https://github.com/pyca/pynacl/commit/b525bd309623437f1e0a2c9e783afcdba91210c5).  Could you supply the build log as lmctv mentioned in the other issue?

ex3me0:
[...]

PR #417, opened 2018-03-24
Also insert build-local lib64 directory in linker search path.

lmctv:
At least in a opensuse42.3 build environment, setup.py install the bundled library in a "lib64" subdirectory, that wasn't present in the linker search path. Should complete #282 and at least solve the problem lamented by @guidow in #406.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -191,6 +191,9 @@ class build_ext(_build_ext):
                 os.path.join(build_clib.build_clib, "include"),
             )
             self.library_dirs.insert(0,
+                os.path.join(build_clib.build_clib, "lib64"),
+            )
+            self.library_dirs.insert(0,
                 os.path.join(build_clib.build_clib, "lib"),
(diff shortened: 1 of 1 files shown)


lmctv:
@guidow, if this works in your environment, please add a confirmation note here.

lmctv:
@guidow; do you think you could try if this change helps?

lmctv:
@guidow, please note that to build the master branch, you need to avoid setting SODIUM_INSTALL=system from the specfile.

lmctv:
@pyca/pynacl-core I've repeated the build error myself, trying to build pynacl in a pristine OpenSuse container, and verified the fix works in the same environment. While it would be nice, I think a confirmation from @guidow is not strictly needed.

lmctv:
@pyca/pynacl-core this PR involves just the build system. Could someone take a review pass?

PR #429, opened 2018-04-27
Matching types of lib.crypto_pwhash binding

pablomartinezm:


lmctv:
@pablomartinezm I don't understand the reason why you are copying the input parameters `password` and `salt` into a couple of temporary buffers, since cffi should already send them in correctly. Further, while you are at least using a different name for the temporary password buffer (I'd prefer something like `pwdbuf` or `passwordbuf`), I really dislike the name rebinding you do at line 395 for the `salt` parameter.

lmctv:
@pablomartinezm would you mind clarifying the problems you experienced with the code as it stands now?

reaperhulk:
Closing due to lack of activity -- if this is still an issue feel free to reopen.

Issue #432, opened 2018-04-30
Add binding to expose crypto_secretbox_macbytes()


lmctv:
Needed for correct use of the nacl.SecretBox and nacl.bindings.crypto_secretbox[…] APIs. See [#431]

lmctv:
Superceded by #433.

PR #433, opened 2018-04-30
Add crypto secretbox macbytes bindings

lmctv:
and some documentation, solving issue #432.


lmctv:
@reaperhulk since there is no change in behavior, I think reviewing this should not take too much effort.

PR #435, opened 2018-05-05
update changelog for libsodium version

slur7:
@alex #433


reaperhulk:
This appears to be a cherry pick of a commit from long ago. We've already released 1.2.0, the changelog has the correct libsodium version, and we're working on 1.3.0.

PR #436, opened 2018-05-30
Update INSTALL.rst

somu1795:
In Raspbian, the compilation fails due to libffi-dev being not available


alex:
Thanks for sending a PR! This should be up at the top of the "source install" section, not in an unrelated warning about easy_install.

reaperhulk:
The added info is great but is in the wrong section. I'm going to close this PR but if you'd still like this information to be added to the docs feel free to reopen after you push a fix!

PR #455, opened 2018-07-14
stop codecov from commenting on PRs

reaperhulk:
we already require 100% coverage, it can just fail the status check


alex:
test failure relevant

PR #471, opened 2018-09-16
Replace use of six.binary_type with bytes

jdufresne:
The type `bytes` exists on all supported Pythons. On Python 2, it is an
alias of `str`, same as six. There is no need to use a compatibility shim.
Makes code more forward compatible with Python 3 idioms.


alex:
@jdufresne Can you please rebase on master? We've fixed the docs test that's failing here.

jdufresne:
Rebased. Thanks.

PR #485, opened 2018-09-28
Update setup_requirements to match pyproject.toml

kubinski:
Updated setup_requirements to match the requirements listed in pyproject.toml.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,8 @@ except ImportError:
 
 
 requirements = ["six"]
-setup_requirements = []
+setup_requirements = ["setuptools",
+                      "wheel"]
 test_requirements = ["pytest>=3.2.1,!=3.3.0",
                      "hypothesis>=3.27.0"]
 docs_requirements = ["sphinx>=1.6.5",


lmctv:
@pyca/pynacl-core @dstufft I seem to remember using setup_requirements is deprecated when building with pip. If this is the case, I think we should add a comment before the empty list in setup.py

reaperhulk:
setup_requires are deprecated by pyproject.toml but we need to retain them until we decide to drop support for pips without pyproject.toml support (aka everything except the latest release right now)

kitterma:
The problem with this change is what wheel is only required if you are building a wheel.  Not everyone does this, so it seems odd to require it.  Particularly since wheel being available to pip is the default in a virtualenv.  Since I'm not building a wheel, now I have to carry this pointless requirement or patch the project.

reaperhulk:
@kitterma Yeah I'd be happy to take a PR removing that. We don't do it in `pyca/cryptography` and there's no reason to do it here.

kitterma:
@reaperhulk OK.  See #596.  Thanks.

PR #488, opened 2018-10-26
Added bindings for crypto_sign_ed25519_sk_to_pk

sfstpala:


src/nacl/bindings/crypto_sign.py
```diff
@@ -165,6 +165,31 @@ def crypto_sign_ed25519_sk_to_curve25519(secret_key_bytes):
     return ffi.buffer(curve_secret_key, curve_secret_key_len)[:]
 
 
+def crypto_sign_ed25519_sk_to_pk(secret_key_bytes):
+    """
+    Converts a secret Ed25519 key (encoded as bytes ``secret_key_bytes``) to
```
lmctv: I'd suggest a bit of rewording here, to put the docstring in line with libsodium's documentation at

https://download.libsodium.org/doc/public-key_cryptography/public-key_signatures#extracting-the-seed-and-the-public-key-from-the-secret-key

which correctly talks about "extraction" rather than "conversion"

src/nacl/bindings/crypto_sign.py
```diff
@@ -165,6 +165,31 @@ def crypto_sign_ed25519_sk_to_curve25519(secret_key_bytes):
...
+    :param public_key_bytes: bytes
+    :rtype: bytes
+    """
+    if len(secret_key_bytes) != crypto_sign_SECRETKEYBYTES:
+        raise exc.ValueError("Invalid secret key")
+
+    public_key_len = crypto_sign_PUBLICKEYBYTES
+    public_key = ffi.new("unsigned char[]", public_key_len)
+
+    rc = lib.crypto_sign_ed25519_sk_to_pk(public_key, secret_key_bytes)
```
lmctv: Sorry for not asking first, but did you check whether this is really faster than 
```
     public_key = secret_key_bytes[crypto_sign_SEEDBYTES:]
```

sfstpala: Yes, good call. It is about 10x faster. I have changed it.

I also added `crypto_sign_ed25519_sk_to_seed`while I was at it.

src/bindings/crypto_sign.h
```diff
@@ -33,10 +33,12 @@ int crypto_sign_open(unsigned char *m,  unsigned long long *mlen,
                const unsigned char *sm, unsigned long long smlen,
                const unsigned char *pk);
 
-
 int crypto_sign_ed25519_sk_to_pk(unsigned char *pk,
```
lmctv: If you verified the direct slicing to be faster than calling libsodium's crypto_sign_ed25519_sk_to_pk(), you should also remove the now unused binding here.

src/bindings/crypto_sign.h
```diff
@@ -33,10 +33,12 @@ int crypto_sign_open(unsigned char *m,  unsigned long long *mlen,
...
                                  const unsigned char *sk);
 
+int crypto_sign_ed25519_sk_to_seed(unsigned char *pk,
```
lmctv: Seems unused as crypto_sign_ed25519_sk_to_pk()

src/bindings/crypto_sign.h
sfstpala: Done. Thanks for being patient as I'm learning my way around the codebase :)

lmctv:
@sfstpala: thank you for your contribution.

PR #498, opened 2018-11-19
Added new bindings

pablomartinezm:
* Added bindings for crypto_stream
* Added bindings for sign and verify detached functions

### DIFF
--- a/src/bindings/crypto_sign.h
+++ b/src/bindings/crypto_sign.h
@@ -61,3 +61,12 @@ int crypto_sign_ed25519ph_final_create(crypto_sign_ed25519ph_state *state,
 int crypto_sign_ed25519ph_final_verify(crypto_sign_ed25519ph_state *state,
                                        unsigned char *sig,
                                        const unsigned char *pk);
+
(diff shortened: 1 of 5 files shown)


lmctv:
@pablomartinezm : there are a some problems with your contribution, which should be solved before merge:
- many source-style problems revealed by flake8, logged at https://travis-ci.org/pyca/pynacl/jobs/457194027
- usually, we stick to the "PR should be independent" rule; I think it would be better if you split out the `crypto_sign_*detached` stuff into a distinct PR
- missing tests for all the added functions, as you can see at https://codecov.io/gh/pyca/pynacl/pull/498/diff

lmctv:
@pablomartinezm would you prefer splitting the PR and adding the needed tests yourself, or you prefer we start with a new PR closing this one?

pablomartinezm:
I'll better close this PR and include all tests in a new one ;)

lmctv:
Just a couple additional notes:
- we usually declare the code as being © Donald Stufft and individual contributors
- I'm unsure about adding low-level bindings for crypto_sign_detached, unless we could benchmark the latter as being overall faster than the slicing we currently use in the high level SigningKey.sign() method.

PR #505, opened 2019-01-16
Revert "Decode smessage (#504)"

lmctv:
This reverts commit 3384aa85e0e495427402130e8dc185f66b16283b.

Full discussion at https://github.com/pyca/pynacl/pull/504

PR #506, opened 2019-01-16
Signature encoders

blag:
This is a follow up to #504, however, this PR is still relevant even if #505 is merged.

This PR adds separate encoders for `smessage` and `signature`, and for `ciphertext` and `nonce`.

Using the previous `encoder` parameter now works, but emits a warning with instructions to use `message_encoder` and `signature_encoder` or `ciphertext_encoder` and `nonce_encoder` instead, as appropriate.

Both the warning and messages are tested.

At-mentioning people here because I think they might be interested (apologies if this is annoying, but I think you will all have some valuable input): @lmctv, @reaperhulk, @alex.

### DIFF
--- a/docs/public.rst
+++ b/docs/public.rst
@@ -219,7 +219,7 @@ Reference
 
         :return: An instance of :class:`~nacl.public.Box`.
 
-    .. method:: encrypt(plaintext, nonce, encoder)
+    .. method:: encrypt(plaintext, nonce, ciphertext_encoder, nonce_encoder)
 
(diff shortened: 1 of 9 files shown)


lmctv:
Since I mostly concur with @reaperhulk desire to get rid of the encoder interface, I don't think adding another encoder to be useful. Would you mind commenting on my proposal to quickly add explicit .raw_* properties to the signed message and, in the long term use the encoder only for the "simple" interface?

PR #514, opened 2019-03-03
Wrap raw blake2 state in a python class.

lmctv:
This way we avoid leaking raw buffers, and can type check all blake2 state usages.

### DIFF
--- a/src/nacl/bindings/crypto_generichash.py
+++ b/src/nacl/bindings/crypto_generichash.py
@@ -14,8 +14,6 @@
 
 from __future__ import absolute_import, division, print_function
 
-import copy
-
 from six import integer_types
 
 from nacl import exceptions as exc
--- a/src/nacl/bindings/crypto_generichash.py
+++ b/src/nacl/bindings/crypto_generichash.py
@@ -14,6 +14,8 @@
 
 from __future__ import absolute_import, division, print_function
 
+import copy
+
 from six import integer_types
 
 from nacl import exceptions as exc
@@ -138,8 +140,6 @@ class Blake2State(object):
         ffi.memmove(_st._statebuf,
                     self._statebuf, crypto_generichash_STATEBYTES)
         return _st
-
-    copy = __copy__
 
 
 def generichash_blake2b_init(key=b'', salt=b'',
--- a/src/nacl/bindings/crypto_generichash.py
+++ b/src/nacl/bindings/crypto_generichash.py
@@ -1,4 +1,4 @@
-# Copyright 2013 Donald Stufft and individual contributors
+# Copyright 2013-2019 Donald Stufft and individual contributors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -122,6 +122,26 @@ def generichash_blake2b_salt_personal(data,
     return ffi.buffer(digest, digest_size)[:]
 
 
+class Blake2State(object):
+    """
+    Python-level wrapper for the crypto_generichash_blake2b state buffer
+    """
+    __slots__ = ['statebuf', 'digest_size']
+
+    def __init__(self, digest_size):
+        self.statebuf = ffi.new("unsigned char[]",
+                                crypto_generichash_STATEBYTES)
+        self.digest_size = digest_size
+
+    def __copy__(self):
+        _st = self.__class__(self.digest_size)
+        ffi.memmove(_st.statebuf,
+                    self.statebuf, crypto_generichash_STATEBYTES)
+        return _st
+
+    copy = __copy__
+
+
 def generichash_blake2b_init(key=b'', salt=b'',
                              person=b'',
                              digest_size=crypto_generichash_BYTES):
@@ -144,13 +164,13 @@ def generichash_blake2b_init(key=b'', salt=b'',
                         the default digest size is
                         :py:data:`.crypto_generichash_BYTES`
     :type digest_size: int
-    :return: an initizialized state buffer
+    :return: a initizialized :py:class:`.Blake2State`
     :rtype: object
     """
 
     _checkparams(digest_size, key, salt, person)
 
-    statebuf = ffi.new("unsigned char[]", crypto_generichash_STATEBYTES)
+    state = Blake2State(digest_size)
 
     # both _salt and _personal must be zero-padded to the correct length
     _salt = ffi.new("unsigned char []", crypto_generichash_SALTBYTES)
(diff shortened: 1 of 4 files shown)


src/nacl/bindings/crypto_generichash.py
```diff
@@ -144,13 +164,13 @@ def generichash_blake2b_init(key=b'', salt=b'',
                         the default digest size is
                         :py:data:`.crypto_generichash_BYTES`
     :type digest_size: int
-    :return: an initizialized state buffer
+    :return: a initizialized :py:class:`.Blake2State`
```
reaperhulk: initialized is misspelled

src/nacl/bindings/crypto_generichash.py
```diff
@@ -122,6 +122,26 @@ def generichash_blake2b_salt_personal(data,
...
 
 
+class Blake2State(object):
+    """
+    Python-level wrapper for the crypto_generichash_blake2b state buffer
+    """
+    __slots__ = ['statebuf', 'digest_size']
+
+    def __init__(self, digest_size):
+        self.statebuf = ffi.new("unsigned char[]",
```
reaperhulk: This should be underscore prefixed since a `Blake2State` instance should be considered opaque outside of the `digest_size`.

src/nacl/bindings/crypto_generichash.py
```diff
@@ -159,58 +179,57 @@ def generichash_blake2b_init(key=b'', salt=b'',
...
            raising=exc.TypeError)
 
-    rc = lib.crypto_generichash_blake2b_update(statebuf, data, len(data))
+    rc = lib.crypto_generichash_blake2b_update(state.statebuf, data, len(data))
     ensure(rc == 0, 'Unexpected failure',
            raising=exc.RuntimeError)
 
 
-def generichash_blake2b_final(statebuf, digest_size):
+def generichash_blake2b_final(state):
```
reaperhulk: Do we consider these bindings public API? If so this is a compatiblity break.

lmctv: Since our public documentation doesn't mention low level bindings, I think we can consider anything nacl.bindings a private API.

src/nacl/bindings/crypto_generichash.py
lmctv: Will change

src/nacl/bindings/crypto_generichash.py
lmctv: At least unti #494 gets in, or someone starts writing explicit docs for low-level bindings :wink:

src/nacl/bindings/crypto_generichash.py
```diff
@@ -122,6 +122,26 @@ def generichash_blake2b_salt_personal(data,
...
+    Python-level wrapper for the crypto_generichash_blake2b state buffer
+    """
+    __slots__ = ['_statebuf', 'digest_size']
+
+    def __init__(self, digest_size):
+        self._statebuf = ffi.new("unsigned char[]",
+                                 crypto_generichash_STATEBYTES)
+        self.digest_size = digest_size
+
+    def __copy__(self):
```
reaperhulk: Since this is copying the state to a new object rather than a reference isn't this more correctly `__deepcopy__`?

src/nacl/bindings/crypto_generichash.py
```diff
@@ -122,6 +122,26 @@ def generichash_blake2b_salt_personal(data,
...
+        _st = self.__class__(self.digest_size)
+        ffi.memmove(_st._statebuf,
+                    self._statebuf, crypto_generichash_STATEBYTES)
+        return _st
+
+    copy = __copy__
```
reaperhulk: Not sure how I feel about this. Implementing `__copy__` implies using `copy.copy` is what you want to support, right (or really `copy.deepcopy` based on my comment above). So is there a reason to alias the method in this fashion rather than `from copy import deepcopy; newst = deepcopy(stateobj)`?

src/nacl/bindings/crypto_generichash.py
lmctv: The correct copy behaviour is for the copy of the state to be a different object, like happens for lists. copy.deepcopy already works correctly.

src/nacl/bindings/crypto_generichash.py
lmctv: The aliasing was not really needed. Removed.

lmctv:
@reaperhulk  Segmentation faults gone. Travis is back to green;

lmctv:
As a matter of facts, I didn't experience any segmentation fault after building _sodium with clang version 7.

src/nacl/bindings/crypto_generichash.py
reaperhulk: The semantics of `__copy__` are normally that you get a new object that contains identical references for members. `__deepcopy__` is a new object and all of its members are also new objects. So this seems like a deepcopy to me?

lmctv: IMO this is just like what happens for mutable types, since in their case copy.copy creates a new instance, with the same initial content:
```
$ python3
Python 3.7.2+ (default, Feb  2 2019, 14:31:48)
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> l1 = [1, 2, 3]
>>> import copy
>>> l2 = copy.copy(l1)
>>> l1 is l2
False
>>> l2.append(4)
>>> l1
[1, 2, 3]
>>> l2
[1, 2, 3, 4]
>>> d1 = {1:1, 2:2}
>>> d2 = copy.copy(d1)
>>> d2 is d1
False
>>> d2[3] = 3
>>> d1
{1: 1, 2: 2}
>>> d2
{1: 1, 2: 2, 3: 3}
>>>
```

lmctv: The reference to a correct behaviour of deepcopy in https://github.com/pyca/pynacl/pull/514#discussion_r262255096 is outdated, since that worked correctly in the aborted bytearray attempt.

reaperhulk: Yes, copy makes a new top level object, but the documentation states that it uses references for the members of that new object. So `_statebuf` would be the same reference, rather than a new object itself. deepcopy, by contrast, also creates copies of all members recursively.

This is a pedantic argument since the implementation here is correct -- I just think we should be using `__deepcopy__` as the name of our method and calling `deepcopy` the function, which will use `__deepcopy__`.

PR #516, opened 2019-03-11
Update secret.rst

LeeSmet:
As described in https://libsodium.gitbook.io/doc/secret-key_cryptography/authenticated_encryption#algorithm-details, libsodium uses XSalsa20 instead of Salsa20 for its underlying encryption, also update the link to point to the libsodium docs about XSalsa20 for reference

### DIFF
--- a/docs/secret.rst
+++ b/docs/secret.rst
@@ -181,5 +181,5 @@ Reference
 Algorithm details
 -----------------
 
-:Encryption: `Salsa20 stream cipher <https://en.wikipedia.org/wiki/Salsa20>`_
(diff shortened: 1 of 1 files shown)


docs/secret.rst
```diff
@@ -181,5 +181,5 @@ Reference
 Algorithm details
 -----------------
 
-:Encryption: `Salsa20 stream cipher <https://en.wikipedia.org/wiki/Salsa20>`_
+:Encryption: `XSalsa20 stream cipher <https://libsodium.gitbook.io/doc/advanced/stream_ciphers/xsalsa20>`_
```
lmctv: The change to xsalsa is correct; while I'm unsure the linked page from libsodium is much better than the Wikipedia one,  any reference to the right algorithm is better than the error we had...

lmctv:
Thanks, @LeeSmet

Issue #517, opened 2019-03-15
Decrypting data with SealedBox(less than 48 bytes?) gives ValueError: negative array length 


ThePrzeglo:
Steps to reproduce:
```
python3.7 -m venv pynacl_test
source pynacl_test/bin/activate
pip install pynacl
python script.py # Code below
```

Code:
```
from nacl.public import PrivateKey, PublicKey, SealedBox
import nacl
import os

def trigger_bug_a():
    k = PrivateKey.generate()
    box = SealedBox(k)
    box.decrypt(b"")

def trigger_bug_b():
    for i in range(1000, 0, -1):
        try:
            k = PrivateKey.generate()
            box = SealedBox(k)
            box.decrypt(os.urandom(i))
        except nacl.exceptions.CryptoError: # Expected exception?
            pass
        except:
            print(f"Raised for i: {i}")
            raise

if __name__ == "__main__":
    trigger_bug_b()

if __name__ == "__main__":
    trigger_bug_b()
```
Result:
```
Raised for i: 47
Traceback (most recent call last):
  File "script.py", line 23, in <module>
    trigger_bug_b()
  File "script.py", line 15, in trigger_bug_b
    box.decrypt(os.urandom(i))
  File "/mnt/c/Users/John/Desktop/pynacl_test/lib/python3.7/site-packages/nacl/public.py", line 361, in decrypt
    self._private_key,
  File "/mnt/c/Users/John/Desktop/pynacl_test/lib/python3.7/site-packages/nacl/bindings/crypto_box.py", line 307, in crypto_box_seal_open
    plaintext = ffi.new("unsigned char[]", _mlen)
ValueError: negative array length
```

Tested on WSL with Python 3.7.1 (CPython)
pip freeze:
```
cffi==1.12.2
pkg-resources==0.0.0
pycparser==2.19
PyNaCl==1.3.0
six==1.12.0
```
I guess throwing exception from ffi is not intended behaviour.

lmctv:
Thank you for the detailed bug report. I agree the exception should get raised *before* even trying to allocate the output buffer.

lmctv:
@reaperhulk @pyca/pynacl-core: do you think we should we backport #518 and release a 1.3.1 version,or we could rely on a exception being already raised on too-short input and just close this issue?

reaperhulk:
I don't think we need to backport. Let's triage a set of issues for a 1.4.0 milestone and see if we can't get that out the door in April though. I'll close this since it's now fixed in master.

PR #518, opened 2019-03-15
Force positive size for plaintext output array 

lmctv:
Solves #517. The tests have been verified as failing before applying the enclosed changes to `src/nacl/bindings/crypto_box.py`

### DIFF
--- a/src/nacl/bindings/crypto_box.py
+++ b/src/nacl/bindings/crypto_box.py
@@ -304,10 +304,11 @@ def crypto_box_seal_open(ciphertext, pk, sk):
     _clen = len(ciphertext)
     _mlen = _clen - crypto_box_SEALBYTES
 
-    plaintext = ffi.new("unsigned char[]", _mlen)
+    # zero-length malloc results are implementation.dependent
+    plaintext = ffi.new("unsigned char[]", max(1, _mlen))
 
     res = lib.crypto_box_seal_open(plaintext, ciphertext, _clen, pk, sk)
     ensure(res == 0, "An error occurred trying to decrypt the message",
            raising=exc.CryptoError)
 
-    return ffi.buffer(plaintext, _mlen)[:]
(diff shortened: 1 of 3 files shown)


lmctv:
Please send any further comments about the original issue #517 to the proposed solution PR #518

src/nacl/bindings/crypto_box.py
```diff
@@ -304,10 +304,11 @@ def crypto_box_seal_open(ciphertext, pk, sk):
     _clen = len(ciphertext)
     _mlen = _clen - crypto_box_SEALBYTES
```
lmctv: I'm unsure about adding a explicit check to raise ValueError when `_mlen < 0`.

lmctv:
@pyca/pynacl-core @reaperhulk I think this is ready for review.

src/nacl/bindings/crypto_box.py
lmctv: to clarify, if we would decide add a explicit size check, IMO the raised exception should be a `nacl.exception.CryptoError` subclass; both `nacl.exception.TypeError`  and `nacl.exception.ValueError` could make sense.

reaperhulk: If we do an explicit check why not just do `_mlen <= 0` and then we don't need to do the `max()` code below. Or is it possible to seal an empty byte string such that when you call crypto_box_seal_open you would expect an empty byte string response?

lmctv: While I didn't add a explicit test for that case, I verified sealedbox box can ship **and verify** a zero-length message while preparing the PR. If you think the test can be useful, I can either add it in this PR, or in a follow-up one.

lmctv: Still, if we add a explicit check for `_mlen < 0`, we can skip the second `max`.

reaperhulk: Let's add the negative test so we can remove that second max

PR #523, opened 2019-04-13
Deprecate encoder in signining

lmctv:
Revert #504 and deprecate usage of the encoder= parameter in high level signing APIs, as suggested by @reaperhulk  in his review of #507.
If we decide to follow this route, I think we should do the same for the other high level APIs too.
Since this is a potentially disruptive change for downstream users, I'd be grateful if also @dstufft could give his opinion on these changes.

### DIFF
--- a/docs/signing.rst
+++ b/docs/signing.rst
@@ -17,16 +17,15 @@ use it to validate that your messages are actually authentic.
 Example
 -------
 
-Signing and verifying a message without encoding the key or message
-
 Signer's perspective (:class:`~nacl.signing.SigningKey`)
 
 .. testcode::
 
-    from nacl.signing import SigningKey
+    import nacl.encoding
+    import nacl.signing
 
     # Generate a new random signing key
-    signing_key = SigningKey.generate()
+    signing_key = nacl.signing.SigningKey.generate()
 
     # Sign a message with the signing key
     signed = signing_key.sign(b"Attack at Dawn")
@@ -35,138 +34,26 @@ Signer's perspective (:class:`~nacl.signing.SigningKey`)
     verify_key = signing_key.verify_key
 
     # Serialize the verify key to send it to a third party
-    verify_key_bytes = verify_key.encode()
+    verify_key_hex = verify_key.encode(encoder=nacl.encoding.HexEncoder)
 
 Verifier's perspective (:class:`~nacl.signing.VerifyKey`)
 
 .. testcode::
 
-    from nacl.signing import VerifyKey
+    import nacl.signing
 
     # Create a VerifyKey object from a hex serialized public key
-    verify_key = VerifyKey(verify_key_bytes)
+    verify_key = nacl.signing.VerifyKey(verify_key_hex,
+                                        encoder=nacl.encoding.HexEncoder)
 
     # Check the validity of a message's signature
-    # The message and the signature can either be passed together, or
-    # separately if the signature is decoded to raw bytes.
-    # These are equivalent:
+    # The message and the signature can either be passed separately or
+    # concatenated together.  These are equivalent:
     verify_key.verify(signed)
     verify_key.verify(signed.message, signed.signature)
 
     # Alter the signed message text
     forged = signed[:-1] + bytes([int(signed[-1]) ^ 1])
-    # Will raise nacl.exceptions.BadSignatureError, since the signature check
-    # is failing
-    verify_key.verify(forged)
-
-.. testoutput::
-
-    Traceback (most recent call last):
-     ...
-    nacl.exceptions.BadSignatureError: Signature was forged or corrupt
-
-
-Example
--------
-
-Signing and verifying a message encoded with HexEncoder
-
-Signer's perspective (:class:`~nacl.signing.SigningKey`)
-
-.. testcode::
-
-    from nacl.encoding import HexEncoder
-    from nacl.signing import SigningKey
-
-    # Generate a new random signing key
-    signing_key = SigningKey.generate()
-
-    # Sign a message with the signing key
-    signed_hex = signing_key.sign(b"Attack at Dawn", encoder=HexEncoder)
-
-    # Obtain the verify key for a given signing key
-    verify_key = signing_key.verify_key
-
-    # Serialize the verify key to send it to a third party
-    verify_key_hex = verify_key.encode(encoder=HexEncoder)
-
-Verifier's perspective (:class:`~nacl.signing.VerifyKey`)
-
-.. testcode::
-
-    from nacl.encoding import HexEncoder
-    from nacl.signing import VerifyKey
-
-    # Create a VerifyKey object from a hex serialized public key
-    verify_key = VerifyKey(verify_key_hex, encoder=HexEncoder)
-
-    # Check the validity of a message's signature
-    # The message and the signature can either be passed together, or
-    # separately if the signature is decoded to raw bytes.
-    # These are equivalent:
-    verify_key.verify(signed_hex, encoder=HexEncoder)
-    signature_bytes = HexEncoder.decode(signed_hex.signature)
-    verify_key.verify(signed_hex.message, signature_bytes,
-                      encoder=HexEncoder)
-
-    # Alter the signed message text
-    forged = signed_hex[:-1] + bytes([int(signed_hex[-1]) ^ 1])
-    # Will raise nacl.exceptions.BadSignatureError, since the signature check
-    # is failing
-    verify_key.verify(forged)
-
-.. testoutput::
-
-    Traceback (most recent call last):
-     ...
-    nacl.exceptions.BadSignatureError: Signature was forged or corrupt
-
-
-Example
--------
-
-Signing and verifying a message encoded with Base64Encoder
-
-Signer's perspective (:class:`~nacl.signing.SigningKey`)
-
-.. testcode::
-
-    from nacl.encoding import Base64Encoder
-    from nacl.signing import SigningKey
-
-    # Generate a new random signing key
-    signing_key = SigningKey.generate()
-
-    # Sign a message with the signing key
-    signed_b64 = signing_key.sign(b"Attack at Dawn", encoder=Base64Encoder)
-
-    # Obtain the verify key for a given signing key
-    verify_key = signing_key.verify_key
-
-    # Serialize the verify key to send it to a third party
-    verify_key_b64 = verify_key.encode(encoder=Base64Encoder)
-
-Verifier's perspective (:class:`~nacl.signing.VerifyKey`)
-
-.. testcode::
-
-    from nacl.encoding import Base64Encoder
-    from nacl.signing import VerifyKey
-
-    # Create a VerifyKey object from a base64 serialized public key
-    verify_key = VerifyKey(verify_key_b64, encoder=Base64Encoder)
-
-    # Check the validity of a message's signature
-    # The message and the signature can either be passed together, or
-    # separately if the signature is decoded to raw bytes.
-    # These are equivalent:
-    verify_key.verify(signed_b64, encoder=Base64Encoder)
-    signature_bytes = Base64Encoder.decode(signed_b64.signature)
-    verify_key.verify(signed_b64.message, signature_bytes,
-                      encoder=Base64Encoder)
-
-    # Alter the signed message text
-    forged = signed_b64[:-1] + bytes([int(signed_b64[-1]) ^ 1])
     # Will raise nacl.exceptions.BadSignatureError, since the signature check
     # is failing
     verify_key.verify(forged)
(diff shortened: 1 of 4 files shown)


lmctv:
@reaperhulk I think it makes sense, if we are willing to "just deal with bytes" in the long term, to just begin deprecating all the encoder= APIs, instead of doing a step just deprecating non-raw encoders.

lmctv:
@dstufft , @reaperhulk , @pyca/pynacl-core: could someone take a look at this to push it forward?

lmctv:
@dstufft @reaperhulk @pyca/pynacl-core ping

src/nacl/signing.py
```diff
@@ -59,7 +61,20 @@ class VerifyKey(encoding.Encodable, StringFixer, object):
...
+        if len(args) == 1:
+            encoder = args[0]
+        elif len(args) > 1:
+            raise TypeError("Too many arguments")
+        else:
+            encoder = kwargs.get('encoder', encoding.RawEncoder)
+
+        if 'encoder' in kwargs or len(args) > 0:
+            warn(("Implicit encoding/decoding of signing keys "
+                  "is deprecated and will get removed in a future release. "
```
reaperhulk: s/will get/will be

src/nacl/signing.py
```diff
@@ -59,7 +61,20 @@ class VerifyKey(encoding.Encodable, StringFixer, object):
     :param encoder: A class that is able to decode the `key`
     """
 
-    def __init__(self, key, encoder=encoding.RawEncoder):
+    def __init__(self, key, *args, **kwargs):
+        if len(args) == 1:
```
reaperhulk: I don't like this approach. Why not just do `encoder=None` and if it's not `None` pass that as the encoder after warning that passing encoder is deprecated?

src/nacl/signing.py
```diff
@@ -101,13 +116,27 @@ def verify(self, smessage, signature=None, encoder=encoding.RawEncoder):
             signature.
         :rtype: :class:`bytes`
         """
+        if len(args) == 1:
```
reaperhulk: Same comment about the approach.

src/nacl/signing.py
```diff
@@ -101,13 +116,27 @@ def verify(self, smessage, signature=None, encoder=encoding.RawEncoder):
...
+            encoder = args[0]
+        elif len(args) > 1:
+            raise TypeError("Too many arguments")
+        else:
+            encoder = kwargs.get('encoder', encoding.RawEncoder)
+
+        if 'encoder' in kwargs or len(args) > 0:
+            warn(("Automatic encoding/decoding of signed messages "
+                  "is deprecated and will get removed in a future release. "
```
reaperhulk: Use "will be" instead of "will get"

src/nacl/signing.py
```diff
@@ -141,7 +170,22 @@ class SigningKey(encoding.Encodable, StringFixer, object):
         (i.e. public) key that corresponds with this signing key.
     """
 
-    def __init__(self, seed, encoder=encoding.RawEncoder):
+    def __init__(self, seed, *args, **kwargs):
+
+        if len(args) == 1:
```
reaperhulk: Same

src/nacl/signing.py
```diff
@@ -141,7 +170,22 @@ class SigningKey(encoding.Encodable, StringFixer, object):
...
+            encoder = args[0]
+        elif len(args) > 1:
+            raise TypeError("Too many arguments")
+        else:
+            encoder = kwargs.get('encoder', encoding.RawEncoder)
+
+        if 'encoder' in kwargs or len(args) > 0:
+            warn(("Implicit encoding/decoding of signing keys "
+                  "is deprecated and will get removed in a future release. "
```
reaperhulk: Same

src/nacl/signing.py
```diff
@@ -182,19 +226,30 @@ def generate(cls):
...
-    def sign(self, message, encoder=encoding.RawEncoder):
+    def sign(self, message, *args, **kwargs):
         """
         Sign a message using this key.
 
         :param message: [:class:`bytes`] The data to be signed.
         :param encoder: A class that is used to encode the signed message.
         :rtype: :class:`~nacl.signing.SignedMessage`
         """
+        if len(args) == 1:
```
reaperhulk: Same

src/nacl/signing.py
```diff
@@ -182,19 +226,30 @@ def generate(cls):
...
+            encoder = args[0]
+        elif len(args) > 1:
+            raise TypeError("Too many arguments")
+        else:
+            encoder = kwargs.get('encoder', encoding.RawEncoder)
+
+        if 'encoder' in kwargs or len(args) > 0:
+            warn(("Automatic encoding/decoding of signed messages "
+                  "is deprecated and will get removed in a future release. "
```
reaperhulk: same

tests/test_signing.py
```diff
@@ -168,38 +201,6 @@ def test_invalid_signed_message(self):
             forged = SignedMessage(signature + message)
             skey.verify_key.verify(forged)
 
-    def test_base64_smessage_with_detached_sig_matches_with_attached_sig(self):
```
reaperhulk: We shouldn't remove these tests until we're ready to remove the encoder entirely. Just `with pytest.raises` on the encoder lines.

src/nacl/signing.py
lmctv: I think, if we follow the simpler approach, we wouldn't warn the user
when she calls either
```
VerifyKey(key, <expected_encoder>)
```
or
```
VerifyKey(key, encoder=<expected_encoder>)
```
If you think this is over-protective, I'll change.

reaperhulk: What is expected encoder in your example? The default arg?

lmctv: Right

tests/test_signing.py
lmctv: Both these test where added in #504 the lines are deleted from the `Revert "Decode smessage" #504` commit.

src/nacl/signing.py
lmctv: Even If we
```
    def __init__(self, key, encoder=None):
```
instead of
```
    def __init__(self, key, encoder=encoding.RawEncoder):
```
We would not warn for calls to `VerifyKey(key,  encoder=None)` or `VerifyKey(key, None)`.

src/nacl/signing.py
lmctv: Fixed by just pushed 7652934908e665625dee8797d274e30d0ee6dae3

src/nacl/signing.py
reaperhulk: I'm not worried about people doing that unless `None` did something in the old API already.

lmctv: The good thing is, if we change the default value for `encoder`, the warning will be emitted for all callers predating the change. I'm just unsure about not emitting the warning when the argument is passed-in with the expected value.

lmctv: OK, missed your previous message. `None` was not doing anything in the old API. Changes upcoming.

src/nacl/signing.py
lmctv: I think a75cb45f35d17529ebc316dc72f8691fa360a537 could go.

docs/signing.rst
```diff
@@ -35,21 +34,21 @@ Signer's perspective (:class:`~nacl.signing.SigningKey`)
     verify_key = signing_key.verify_key
 
     # Serialize the verify key to send it to a third party
-    verify_key_bytes = verify_key.encode()
+    verify_key_hex = nacl.encoding.HexEncoder.encode(bytes(verify_key))
```
reaperhulk: Seems like we should deprecate `nacl.encoding` entirely? That package only provides interfaces that can be easily replaced by stdlib calls.

lmctv: Yes, but I'd prefer going step by step:
 - signing (almost there?)
 - public
 - secret
 - hash
 - pwhash
 - encoding

lmctv: I think a tracing issue for nacl.encoding deprecation would be useful.

reaperhulk: Okay, go ahead and file the issue and we can track against that for these items.

lmctv: Opened #532

tests/test_signing.py
```diff
@@ -78,30 +79,65 @@ def test_different_keys_are_not_equal(self, k2):
...
+            sk.sign(unsigned,
+                    encoder=HexEncoder)
+        with pytest.warns(PyNaclDeprecated):
+            sk.sign(unsigned,
+                    HexEncoder)
+
+    @pytest.mark.parametrize("hseed", [
+        b"77076d0a7318a57d3c16c17251b26645df4c2f87ebc0992ab177fba51db92c2a",
+    ])
+    def test_raising_on_excess_encoder_parameter(self, hseed):
```
reaperhulk: These tests aren't really testing anything interesting with the new approach, correct?

lmctv: Right, they were redundant. Reverted the relevant commit.

lmctv:
@reaperhulk I hope this is now ready for another review.

lmctv:
@reaperhulk ping.

lmctv:
@reaperhulk Would you mind another review cycle and possibly merge this? I'd rather avoid preparing the `encoder=` pre-deprecation warning PR on un-reverted #504 signing documentation.

lmctv:
@reaperhulk could you please take a look here and at the proposed plan for a quick release I outlined in  https://github.com/pyca/pynacl/issues/532#issuecomment-499271044

src/nacl/signing.py
```diff
@@ -56,10 +58,18 @@ class VerifyKey(encoding.Encodable, StringFixer, object):
...
-    :param encoder: A class that is able to decode the `key`
+    :param encoder: (Deprecated) A class that is able to decode the `key`
     """
 
-    def __init__(self, key, encoder=encoding.RawEncoder):
+    def __init__(self, key, encoder=None):
+        if encoder is None:
+            encoder = encoding.RawEncoder
+        else:
+            warn(("Implicit encoding/decoding of signing keys "
```
reaperhulk: We should just say encoding/decoding of signing keys is deprecated. Remove the words explicit and implicit from this msg.

src/nacl/signing.py
```diff
@@ -97,17 +107,26 @@ def verify(self, smessage, signature=None, encoder=encoding.RawEncoder):
...
-        :param encoder: A class that is able to decode the secret message and
-            signature.
+        :param encoder: (Deprecated) A class that is able to decode the secret
+            message and signature.
         :rtype: :class:`bytes`
         """
+        if encoder is None:
+            encoder = encoding.RawEncoder
+        else:
+            warn(("Automatic encoding/decoding of signed messages "
```
reaperhulk: This should be the same message (or better yet, this should be a function that everything invokes)

src/nacl/signing.py
```diff
@@ -97,17 +107,26 @@ def verify(self, smessage, signature=None, encoder=encoding.RawEncoder):
...
+                  "is deprecated and will be removed in a future release. "
+                  "Remove explicit 'encoder={}' calling parameter").format(
+                      encoder.__class__),
+                 PyNaclDeprecated)
+
         if signature is not None:
             # If we were given the message and signature separately, combine
             #   them.
-            smessage = signature + encoder.decode(smessage)
```
reaperhulk: Can we quickly pull this revert into its own PR so we can land that?

src/nacl/signing.py
```diff
@@ -135,13 +154,22 @@ class SigningKey(encoding.Encodable, StringFixer, object):
...
     :ivar: verify_key: [:class:`~nacl.signing.VerifyKey`] The verify
         (i.e. public) key that corresponds with this signing key.
     """
 
-    def __init__(self, seed, encoder=encoding.RawEncoder):
+    def __init__(self, seed, encoder=None):
+        if encoder is None:
+            encoder = encoding.RawEncoder
+        else:
+            warn(("Implicit encoding/decoding of signing keys "
```
reaperhulk: Same

src/nacl/signing.py
```diff
@@ -182,19 +210,26 @@ def generate(cls):
...
         :param message: [:class:`bytes`] The data to be signed.
-        :param encoder: A class that is used to encode the signed message.
+        :param encoder: (Deprecated) A class that is used to encode
+            the signed message.
         :rtype: :class:`~nacl.signing.SignedMessage`
         """
+        if encoder is None:
+            encoder = encoding.RawEncoder
+        else:
+            warn(("Automatic encoding/decoding of signed messages "
```
reaperhulk: Same

src/nacl/signing.py
lmctv: I just reopened the Revert PR #505, and rebased it on current master.

lmctv: For some reason, #505 was kept in the Closed PR list. Just created #545 from the same branch.

reaperhulk:
We should still deprecate encoder but we can open a PR for it at some point.

PR #527, opened 2019-05-10
Update bundled version of libsodium sources to version 1.0.17

dciabrin:
+ commit db6f43d2 from post 1.0.17

This version exposes new low-level ed25519 API, such as non-clamped versions of
curve point multiplication and modular arithmetic operations.

This update also includes commit db6f43d2 to expose modular multiplication,
which is the only modular operation that didn't make it to 1.0.17. This commit
is applied with patch -t -p1, to only include ed25519 modular multiplication,
and drop ristretto255 which is not part of 1.0.17.


lmctv:
I'd rather avoid upgrading our bundled libsodium to a release version + a cherry-picked commit , especially since this would then introduce problems to any downstream user who needs/prefers to compile pynacl against an external shared libsodium library. Waiting for @pyca/pynacl-core  and @reaperhulk opinions before closing.

reaperhulk:
I am in favor of upgrading to the latest release, but yeah I’m not a fan of grabbing an unreleased commit. Has anyone asked when 1.0.18 is going to be released?

dciabrin:
I haven't asked yet, will do that shortly.
Thanks for your feedbacks :)

dciabrin:
Until the modular multiplication is available in 1.0.18, I may be able to use python modulo for my use case, so I could change this PR to update to libsodium 1.0.17 only.
I'd appreciate if we could keep that PR open for a couple of days until I come up with an update.

dciabrin:
OK I'm closing this PR in favour of #533 which only updates to 1.0.17. I'll rebase #528 shortly on top of #533.
I think #533 and soon-to-be-rebased #528 are enough for my use case as I can replicate the missing scalar_mul with python long integers.

PR #528, opened 2019-05-10
 Expose new low-level ed25519 arithmetic and point multiplication

dciabrin:
This PR depends on #527, it provides bindings to the new low-level ed25519 API that ship with libsodium 1.0.17 and forward: non-clamped curve point multiplication, and arithmetic modulo the order of the main subgroup.

### DIFF
--- a/src/nacl/bindings/crypto_core.py
+++ b/src/nacl/bindings/crypto_core.py
@@ -20,6 +20,9 @@ from nacl.exceptions import ensure
 
 
 crypto_core_ed25519_BYTES = lib.crypto_core_ed25519_bytes()
+crypto_core_ed25519_SCALARBYTES = lib.crypto_core_ed25519_scalarbytes()
+crypto_core_ed25519_NONREDUCEDSCALARBYTES = \
+    lib.crypto_core_ed25519_nonreducedscalarbytes()
 
 
 def crypto_core_ed25519_is_valid_point(p):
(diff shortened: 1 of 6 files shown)


lmctv:
Since the added bindings break travis ci, since the tarball at https://download.libsodium.org/libsodium/releases/LATEST.tar.gz is at version 1.0.17, I'd suggest we either:
 1. upgrade the bundled libsodium to 1.0.17, add only the bindings to what's available therein, and postpone the bindings for crypto_core_ed25519_scalar_mul 
 2. just wait to upgrade the bundled version to a proper release of 1.0.18 and then add these proposed bindings.

dciabrin:
Until the crypto_core_ed25519_scalar_mul is available in 1.0.18, I may be able to use python modulo for my use case, so I could change this PR to update to only provide bindings to the new ed25519 arithmetic in 1.0.17.
I'd appreciate if we could keep that PR open for a couple of days until I come up with an update.

reaperhulk:
1.0.18 is now out! tracking issue to merge the update is https://github.com/pyca/pynacl/issues/536 and the azure pipelines migration already includes 1.0.18 for windows!

dciabrin:
> 1.0.18 is now out! tracking issue to merge the update is #536 and the azure pipelines migration already includes 1.0.18 for windows!

Awesome, are you preparing a PR for it or do you want me to provide one?

lmctv:
@dciabrin if you don't mind, I'd happily review and hopefully merge a 1.0.18, leaving @reaperhulk's  attention to  the azure pipeline update PR.

dciabrin:
@lmctv ack I'll prepare one shortly then, thanks.

lmctv:
@dciabrin I think now's the right time for rebasing this PR!

dciabrin:
OK I just rebased and got rid of the update in CHANGELOG.rst as it's essentially what's written in #543.

Also I purposely didn't add bindings for crypto_core_ed25519_{random,from_hash} as it seems that a similar crypto_core_ed25519_from_uniform was temporary removed a while ago due to a lack of unit test? If need be I can provide another PR for those ones and replicate the testing logic from https://github.com/jedisct1/libsodium/blob/master/test/default/core_ed25519.c#L59.

src/nacl/bindings/crypto_core.py
```diff
@@ -102,3 +105,197 @@ def crypto_core_ed25519_sub(p, q):
...
+    r = ffi.new("unsigned char[]", crypto_core_ed25519_SCALARBYTES)
+
+    lib.crypto_core_ed25519_scalar_complement(r, s)
+
+    return ffi.buffer(r, crypto_core_ed25519_SCALARBYTES)[:]
+
+
+def crypto_core_ed25519_scalar_add(p, q):
+    """
+    Add integers ``p`` and ``p`` modulo ``L``, where ``L`` is the order of
```
lmctv: Add integers ``p`` and ``q``

src/nacl/bindings/crypto_core.py
```diff
@@ -102,3 +105,197 @@ def crypto_core_ed25519_sub(p, q):
...
+    r = ffi.new("unsigned char[]", crypto_core_ed25519_SCALARBYTES)
+
+    lib.crypto_core_ed25519_scalar_sub(r, p, q)
+
+    return ffi.buffer(r, crypto_core_ed25519_SCALARBYTES)[:]
+
+
+def crypto_core_ed25519_scalar_mul(p, q):
+    """
+    Multiply integers ``p`` and ``p`` modulo ``L``, where ``L`` is the
```
lmctv: ``p`` and ``q``

src/nacl/bindings/crypto_core.py
dciabrin: oops, thanks, fixed

src/nacl/bindings/crypto_core.py
dciabrin: fixed

tests/test_bindings.py
```diff
@@ -576,3 +576,121 @@ def test_scalarmult_ed25519_base():
     p2 = c.crypto_scalarmult_ed25519(sclr, BASEPOINT)
 
     assert p2 == p
+
+
+def test_scalarmult_ed25519_noclamp():
+    one = 32 * b'\x01'
```
lmctv: I don't really like calling this big integer `one`, especially since this differs from the following `one`s.

dciabrin: Yes agreed... the intent was to use a value which is known to differ once clamped.
Would it work if I replace `one` and `one_clamped` with resp. `scalar` and `scalar_clamped` ?

```
def test_scalarmult_ed25519_noclamp():
    # An arbitrary scalar which is known to differ once clamped
    scalar = 32 * b'\x01'
[...]
```

lmctv: Anything replacing `one` goes! OK for `scalar`

lmctv:
Thank you @dciabrin

PR #533, opened 2019-05-20
Update bundled version of libsodium sources to version 1.0.17

dciabrin:
This version exposes new low-level ed25519 API, such as non-clamped versions of
curve point multiplication and modular arithmetic operations.


reaperhulk:
Note to self: Independently of this I'll need to upgrade the compiled version on the windows builders.

dciabrin:
> Please add those missing files, and we should be good to go.

Oops, I wasn't sure about those files... I added them back and updated .gitignore to whitelist them for the future updates.

lmctv:
PR checked:
 - downloaded https://download.libsodium.org/libsodium/releases/libsodium-1.0.17.tar.gz
 - verified its signature via the minisign tool
 - extracted the contents of the downloaded tar file
 - verified the contents of src/libsodium is exactly equal to the contents of the libsodium-1.0.17 extracted from the tar file.
@dciabrin: please don't rebase #528 on master until @reaperhulk  closes #534

PR #535, opened 2019-05-26
azure support

reaperhulk:


.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: python
 python: 2.7
-group: edge
-dist: trusty
 sudo: required
```
alex: You can remove this too while you're here.

azure-pipelines.yml
```diff
@@ -0,0 +1,131 @@
...
+        PYTHON_DIR: 'Python37'
+        SODIUM_LIB_PATH: 'C:/libsodium/x64/Release/v140/static'
+  steps:
+  - powershell: |
+      $url = "https://download.libsodium.org/libsodium/releases/libsodium-1.0.17-msvc.zip"
+      Write-Host ('Downloading {0} ...' -f $url)
+      [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+      Invoke-WebRequest -Uri $url -OutFile 'libsodium.zip'
+      Expand-Archive libsodium.zip -DestinationPath c:\libsodium
+    displayName: 'Download libsodium libraries'
```
alex: Why do we do this instead of just using the in-tree copy?

PR #541, opened 2019-06-01
Update bundled version of libsodium sources to version 1.0.18 (#536)

dciabrin:
This version extends the ed25519 API, with scalar*scalar (mod L)
multiplication, and new functions to map a hash to an edwards25519
point or get a random point.  It also implements finite field
arithmetic for the Ristretto group.


lmctv:
@dciabrin sorry, but ot seems this time there is more stuff in the committed libsodium than in the upstream tarball:
```
$ diff -r -u ../../../libsodium-1.0.18 ./
Only in ./: .gitignore
Only in ./: .travis.yml
Only in ./: appveyor.yml
Only in ./: azure-pipelines.yml
Only in ./builds: .gitignore
Only in ../../../libsodium-1.0.18: compile
Only in ../../../libsodium-1.0.18: depcomp
Only in ./dist-build: emscripten-symbols.def
Only in ./dist-build: generate-emscripten-symbols.sh
Only in ./dist-build: osx.sh
Only in ../../../libsodium-1.0.18: install-sh
Only in ./: lgtm.yml
Only in ./: logo.png
Only in ../../../libsodium-1.0.18: ltmain.sh
Only in ./m4: ax_check_gnu_make.m4
Only in ../../../libsodium-1.0.18: missing
Only in ./: packaging
Only in ./: regen-msvc
Only in ./test: constcheck.sh
Only in ./test/default: index.html.tpl
```
Are you sure you downloaded and extracted the right tarball from https://download.libsodium.org/libsodium/releases/ ?

dciabrin:
@lmctv argh apologies I've downloaded https://github.com/jedisct1/libsodium/archive/1.0.18-RELEASE.tar.gz instead. Let me fix that at once...

dciabrin:
Also a quick note before I make you lose more time: libsodium-1.0.18 doesn't include config.guess and config.sub anymore on the root dir. They now include it under ./build-aux sub directory.

Unfortunately due to https://github.com/jedisct1/libsodium/commit/515d540524619987443d88c097888ee74d3fd142, those two files are now bundled without +x permission.

setup.py ensures they get +x, but as a side effect they now show as modified in git as soon as setup.py is ran.
I don't know what we want to do with that one?

setup.py
```diff
@@ -138,12 +138,12 @@ def run(self):
...
 
-        # Ensure all of our executanle files have their permission set
+        # Ensure all of our executable files have their permission set
         for filename in [
                 "src/libsodium/autogen.sh",
                 "src/libsodium/compile",
-                "src/libsodium/config.guess",
-                "src/libsodium/config.sub",
+                "src/libsodium/build-aux/config.guess",
+                "src/libsodium/build-aux/config.sub",
```
lmctv: By looking at how the commands involving `config.guess` and `config.sub` are invoked in ./configure:
```
`$SHELL "$ac_aux_dir/config.guess"`
`$SHELL "$ac_aux_dir/config.sub"`
```
it seems the execute permission is not needed anymore for the configure step to succeed.
I'd suggest to just remove these files from the list, and verify the library is still correctly built.

lmctv: After checking the boundled version is equal to the one extracted from a pristine upstream tar, I just verified the git worktree stays clean after building. Merging.

PR #545, opened 2019-06-23
Revert "Decode smessage (#504)"

lmctv:
This reverts commit 3384aa85e0e495427402130e8dc185f66b16283b.

Full discussion at https://github.com/pyca/pynacl/pull/504

### DIFF
--- a/docs/signing.rst
+++ b/docs/signing.rst
@@ -17,16 +17,15 @@ use it to validate that your messages are actually authentic.
 Example
 -------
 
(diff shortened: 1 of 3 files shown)


lmctv:
@reaperhulk  weekly ping.

reaperhulk:
...so we (I) sat on this so long I'm now wondering if we should revert. We forced a BC incompatible change but now it's been in the wild for coming up on a year...probably best not to flip it back.

Issue #402, opened 2018-02-04
Can't resolve symbol under ARM uClibc


MrDini123:
Hi,

I have an ARMv5 NAS with uClibc C compiler, and Python 3.5.1. I installed pynacl with pip first, and it gave me this:

```
>>> import nacl.secret

/bin/python3.5: symbol '__stack_chk_guard': can't resolve symbol

/bin/python3.5: symbol '__stack_chk_fail': can't resolve symbol
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/lib/python3.5/site-packages/nacl/secret.py", line 17, in <module>
    import nacl.bindings
  File "/lib/python3.5/site-packages/nacl/bindings/__init__.py", line 17, in <module>
    from nacl.bindings.crypto_aead import (
  File "/lib/python3.5/site-packages/nacl/bindings/crypto_aead.py", line 18, in <module>
    from nacl._sodium import ffi, lib
ImportError: unknown dlopen() error
```

I also tried to compile manually the latest git version, but the results are the same. Can you please fix it by checking that in the configure script?

Thanks!

MrDini123:
Nevermind. It was a libsodium bug. I modified the setup.py, and manually added "--disable-ssp" to the configure flags. I hope it will work now.

PR #558, opened 2019-10-15
Fix build on Solaris

### LINKED ISSUE #502: Solaris: import pynacl.bindings fails with "referenced symbol not found"
I'm having trouble getting PyNaCl to work on Solaris. After installation, import fails to load the shared object `_sodium.so` with the message "symbol sodium_unpad: referenced symbol not found".
I am using the "embedded" libsodium, as I won't be able to install libsodium itself on my target machines.

I reproduces this issue on Solaris 10.11 and Solaris 11.2, on both SPARC and x64 architectures, with or without the "-m32" CFLAG (during installation), and on Python 2 and 3.

```
root@host-ci037 (SunOS 5.11) ➜  ~  pip install pynacl
Looking in indexes: https://pypi.org/simple/
Collecting pynacl
  Downloading https://files.pythonhosted.org/packages/61/ab/2ac6dea8489fa713e2b4c6c5b549cc962dd4a842b5998d9e80cf8440b7cd/PyNaCl-1.3.0.tar.gz (3.4MB)
    100% |████████████████████████████████| 3.4MB 663kB/s
  Installing build dependencies ... done
Requirement already satisfied: six in ./python/lib64/python2.7/site-packages/six-1.12.0-py2.7.egg (from pynacl) (1.12.0)
Requirement already satisfied: cffi>=1.4.1 in ./python/lib64/python2.7/site-packages/cffi-1.11.5-py2.7-solaris-2.11-i86pc.64bit.egg (from pynacl) (1.11.5)
Requirement already satisfied: pycparser in ./python/lib64/python2.7/site-packages/pycparser-2.19-py2.7.egg (from cffi>=1.4.1->pynacl) (2.19)
Installing collected packages: pynacl
  Running setup.py install for pynacl ... done
Successfully installed pynacl-1.3.0


root@host-ci037 (SunOS 5.11) ➜  ~  python -c "import nacl.bindings"
  File "<string>", line 1, in <module>
  File "/root/python/lib64/python2.7/site-packages/nacl/bindings/__init__.py", line 17, in <module>
    from nacl.bindings.crypto_aead import (
  File "/root/python/lib64/python2.7/site-packages/nacl/bindings/crypto_aead.py", line 18, in <module>
    from nacl._sodium import ffi, lib
ImportError: ld.so.1: python2.7: fatal: relocation error: file /root/python/lib64/python2.7/site-packages/nacl/_sodium.so: symbol sodium_unpad: referenced symbol not found
```

Any ideas?
Thanks

wiggin15:
This pull request fixes #502 (on Solaris, importing `nacl.bindings` fails with linker errors after building pynacl).
The changes handle two separate issues found:
1. The extension _sodium.so compiles with the flag `-m64` and the rest of the options that come from sysconfig, but libsodium itself does not use the same build configuration when compiling, so it is built as a 32-bit shared object which fails to link. To fix this I am including the build environment variables from sysconfig (those are used to build `_sodium.so`) to the environment variables used when building libsodium.
2. The second problem fixes the linking errors related to the stack guard:
`ImportError: ld.so.1: python2.7: fatal: relocation error: file /root/python/lib64/python2.7/site-packages/PyNaCl-1.3.0-py2.7-solaris-2.11-i86pc.64bit.egg/nacl/_sodium.so: symbol __stack_chk_guard: referenced symbol not found`.
This issue was also reported in #402 and #467.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ import os.path
 import platform
 import subprocess
 import sys
+from distutils.sysconfig import get_config_vars
 
 from setuptools import Distribution, setup
 from setuptools.command.build_ext import build_ext as _build_ext
@@ -129,9 +130,15 @@ class build_clib(_build_clib):
         if use_system():
             return
 
-        build_temp = os.path.abspath(self.build_temp)
+        # use Python's build environment variables
+        build_env = {key: val for key, val in
+                     get_config_vars().items() if key in
+                     ("LDFLAGS", "CFLAGS", "CC", "CCSHARED", "LDSHARED") and
+                     key not in os.environ}
+        os.environ.update(build_env)
 
         # Ensure our temporary build directory exists
+        build_temp = os.path.abspath(self.build_temp)
         try:
             os.makedirs(build_temp)
         except OSError as e:
--- a/setup.py
+++ b/setup.py
@@ -164,12 +164,16 @@ class build_clib(_build_clib):
         configure = abshere("src/libsodium/configure")
 
         # Run ./configure
+        configure_flags = ["--disable-shared", "--enable-static",
+                           "--disable-debug", "--disable-dependency-tracking",
+                           "--with-pic"]
+        if platform.system() == "SunOS":
+            # On Solaris, libssp doesn't link statically and causes linker
+            # errors during import
+            configure_flags.append("--disable-ssp")
         subprocess.check_call(
-            [
-                configure, "--disable-shared", "--enable-static",
-                "--disable-debug", "--disable-dependency-tracking",
-                "--with-pic", "--prefix", os.path.abspath(self.build_clib),
-            ],
+            [configure] + configure_flags +
+            ["--prefix", os.path.abspath(self.build_clib)],
             cwd=build_temp,
         )
 


setup.py
```diff
@@ -161,7 +167,8 @@ def run(self):
             [
                 configure, "--disable-shared", "--enable-static",
                 "--disable-debug", "--disable-dependency-tracking",
-                "--with-pic", "--prefix", os.path.abspath(self.build_clib),
+                "--disable-ssp", "--with-pic",
```
reaperhulk: Passing `--disable-ssp` unconditionally removes a security feature that works in most contexts. If we need to disable it on some platforms we need to be smarter about how we detect support.

setup.py
```diff
@@ -129,9 +130,14 @@ def run(self):
         if use_system():
             return
 
-        build_temp = os.path.abspath(self.build_temp)
+        # use Python's build environment variables
+        build_env = {key: val for key, val in get_config_vars().items() if key in
+                     ("LDFLAGS", "CFLAGS", "CC", "CCSHARED", "LDSHARED")
+                     and key not in os.environ}
+        os.environ.update(build_env)
```
reaperhulk: I'm confused why this would be necessary. setuptools/distutils use the build flags of the compiled Python by default so whatever was used during compilation *should* be passed to the compiler/linker when building an extension. It seems you're not seeing that behavior? Am I mistaken in how setuptools works?

wiggin15: setuptools/distutils do use these build flags for compiling the _extension_ (_sodium.so) but setuptools/distutils are not the ones building the _external library_ (libsodium.so). pynacl is compiling libsodium by calling configure+make directly in this function and it is not using setuptools/distutils for that compilation. When we call configure+make directly (`subprocess.check_call` a few lines down) the build environment is simply inherited from the user environment and it is not using the setuptools/distutils environments. This means _sodium.so, which uses libsodium.so, may be built with different flags and thus fail to link (`-m64` in the case of #502 and #467).

reaperhulk: Ah, yes, of course. My mistake!

setup.py
wiggin15: I investigated this a little further and found that the issue with SSP is caused by non-static linking of this library into libsodium. Since libsodium uses libssp (unless --disable-ssp is specified) but doesn't statically link to it on Solaris, we have to add dynamic linking to this library in pynacl itself (`-lssp`, or, add `ssp` to `libraries` in `bindings/build.py`). However, we can't be sure that libssp is really used (e.g. if the package, CSWlibssp0, is not installed - libsodium's `configure` checks SSP's availability), so we don't want to blindly add libssp to the linker. Which brings me back to the simple solution of just using `--disable-ssp` on Solaris.

I saw a few comments that see this as a bug on libsodium's side (and even a forks that fixes this in their configure script - https://git.luxing.im/hlx98007/shadowsocksr-libev/commit/5d83a5ee7fb8afb883ff51a3e75e0e0a5affefd9) but I don't know if libsodium will accept this upstream.

Do you think it's ok to add checks specifically for Solaris in setup.py to add this flag if we detect this operating system?

wiggin15:
I updated the PR, now the flag is added only in Solaris.

PR #561, opened 2019-10-20
Run tests on py38 on macOS

alex:


alex:
OBE - migrating to GA

PR #575, opened 2020-01-28
Fix #547

### LINKED ISSUE #547: crypto_secretstream: different handling of zero-length data in push/pull
#### Problem:

crypto_secretstream_xchacha20poly1305_push allows to encrypt zero-length strings, while corresponding pull function fails to decrypt them with `nacl.exceptions.ValueError: Ciphertext is too short` error.

#### Expected result:

Anything encrypted with crypto_secretstream_xchacha20poly1305_push can then be decrypted with pull function, provided that symmetric key and data is correct.

I.e. either error will be raised on zero-length encryption attempt, or no error raised on decryption, compared to how it currently works.

#### Actual result:

One-way encryption or silent data corruption - push encrypts without any issues, but subsequent decryption always fails, leaving any encrypted data inaccessible.

#### How to reproduce:

Run example code linked by @mmerickel in #345 currently available at this URL:
https://gist.github.com/mmerickel/aee97620e92f4d73bb3e2ea297e7e8b7
...but with data size that aligns exactly with used block size, resulting in last data block being zero-length.

E.g. by running nacl_streams.py script from there like this:
`rm -f foo && dd if=/dev/urandom of=foo bs=20K count=1 && python nacl_streams.py foo`

Script runs encrypt + decrypt round-trip, which results in `nacl.exceptions.ValueError: Ciphertext is too short` error on decryption step, with expected result of successul decryption.

Tested with pynacl 1.3.0 installed from ArchLinux repos, commit log doesn't seem to mention any relevant changes since then, unless I missed them.

fhriley:


reaperhulk:
Looks like a test needs an update here assuming this change is correct.

fhriley:
Fixed the test.

reaperhulk:
This LGTM, but is unfortunately blocked on fixing our catastrophe of a windows CI. Hopefully soon.

PR #596, opened 2020-05-26
Remove spurious requirement for wheel

kitterma:
As discussed in https://github.com/pyca/pynacl/pull/485#issuecomment-634272864

Since #485 was about making setup.py match pyproject.toml, I changed it there too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,7 @@
 # Must be kept in sync with `setup_requirements` in `setup.py`
 requires = [
     "setuptools>=40.8.0",
+    "wheel",
     "cffi>=1.4.1; python_implementation != 'PyPy'",
 ]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -2,7 +2,6 @@
 # Must be kept in sync with `setup_requirements` in `setup.py`
 requires = [
     "setuptools>=40.8.0",
-    "wheel",
```
reaperhulk, line 5: Not withstanding the comments in the old PR, as I understand PEP 517/518 this value needs to stay in the requires for the build-backend (see: https://www.python.org/dev/peps/pep-0518/).

The de-synchronization is an unfortunate consequence, but why should Python packaging be consistent now 😄

kitterma:
Thanks.  I'll put it back.

kitterma:
How's that?

reaperhulk:
Thanks!

PR #602, opened 2020-07-08
Doc: add a "Faster wheel build" section in INSTALL.rst

Lucas-C:
Reference: https://github.com/pyca/pynacl/issues/302

Preview: https://github.com/Lucas-C/pynacl/blob/patch-1/INSTALL.rst


reaperhulk:
Thanks, this is a fine suggestion for people who understand what the `-j` flag does :)

Issue #601, opened 2020-07-02
Wheel support for aarch64


odidev:
**Summary**
Installing pynacl on aarch64 via pip using command "pip3 install pynacl " tries to build wheel from source code

**Problem description**
pynacl don't have wheel for aarch64 on PyPI repository. So, while installing pynacl via pip on aarch64, pip builds wheel for same resulting in it takes more time to install pynacl. Making wheel available for aarch64 will benefit aarch64 users by minimizing pynacl installation time.

**Expected Output**
Pip should be able to download pynacl wheel from PyPI repository rather than building it from source code.

@pynacl-team, please let me know if I can help you building wheel/uploading to PyPI repository. I am curious to make pynacl wheel available for aarch64. It will be a great opportunity for me to work with you.

brubbel:
+1 Same for armhf (raspberry pi). Latest versions of docker-compose have pynacl in the requirements.

odidev:
Ping @reaperhulk, can you please let me know your interest in releasing AArch64 wheel?

odidev:
Gentele Reminder !!!

Lukejo2:
+1 This takes 45 minutes to build while cross-compiling to aarch64 from a gitlab runner.

odidev:
Hi @reaperhulk 

Thanks for reviewing and merging the PR.
Please let me know if any help is required from our side to add wheel support for aarch64.

Thanks

reaperhulk:
The wheel building is the harder problem here.

Right now we use GHA to build wheels, download them, and then upload them. We don't store API tokens capable of uploading wheels with GH since we'd like to limit scope of compromise (although this could potentially be changed if there was sufficient reason).

However, GHA doesn't have AArch64, so we can't build wheels there at all unless we do some kind of qemu hack. And we can't arbitrarily trigger TravisCI jobs like we can GH actions (nor does it have storage -- all Travis style wheel generation needs to build and upload in one step), so it would be difficult to use Travis.

Dish1306:
Hi @reaperhulk ,

We will look how qemu hack can be used on GHA for AArch64 and update you the steps, 
Will that be helpful ??

Thanks

odidev:
Hi @reaperhulk ,

We will look how qemu hack can be used on GHA for aatch64 and update you the steps, 
Will that be helpful?

Thanks

FlyingThunder:
[...]

PR #612, opened 2020-08-05
ARM64 Support to Github Actions for wheel development

odidev:
Added ARM64 Support to Github Actions for wheel development.
Resolves the second step for #601 

@reaperhulk Please review and let me know if any changes required.

### DIFF
--- a/.github/workflows/wheel-builder.yml
+++ b/.github/workflows/wheel-builder.yml
@@ -154,3 +154,37 @@ jobs:
         with:
           name: "pynacl-${{ github.event.client_payload.BUILD_VERSION }}-win-${{ matrix.WINDOWS.ARCH }}-${{ matrix.PYTHON.VERSION }}"
           path: pynacl-wheelhouse\
+
+  manylinux-arm64:
+     name: "Python ${{ matrix.PYTHON }} for manylinux2014-aarch64"
+     runs-on: ubuntu-latest
+     strategy:
+       matrix:
+         PYTHON: ["cp35-cp35m"]
+     steps:
+     - uses: actions/checkout@v2
+     - run: |
+         docker run --rm --privileged hypriot/qemu-register
+     - uses: docker://quay.io/pypa/manylinux2014_aarch64
+       with:
+        args: |
+              bash -c "set -xe;
+              /opt/python/${{ matrix.PYTHON }}/bin/pip install virtualenv;
+              /opt/python/${{ matrix.PYTHON }}/bin/python -m virtualenv .venv;
+              .venv/bin/pip install setuptools wheel cffi six;
+              .venv/bin/pip install -U pip==18.0.0; # downgrade pip for reasons we can't remember but are definitely needed
+              REGEX='cp3([0-9])*';
+              if [[ ${{ matrix.PYTHON }} =~ $REGEX ]]; then
+                  PY_LIMITED_API=\"--build-option --py-limited-api=cp3${BASH_REMATCH[1]}\";
+              fi;
+              .venv/bin/pip wheel pynacl --no-binary pynacl --no-deps --wheel-dir=tmpwheelhouse $PY_LIMITED_API;
+              auditwheel repair tmpwheelhouse/PyNaCl*.whl -w wheelhouse/;
+              .venv/bin/pip install -U pip; # upgrade so it knows how to install a manylinux2014 wheel, sigh
+              .venv/bin/pip install pynacl --no-index -f wheelhouse/;
+              .venv/bin/python -c \"import nacl.signing; key = nacl.signing.SigningKey.generate();signature = key.sign(b'test'); key.verify_key.verify(signature)\";"
+     - run: mkdir pynacl-wheelhouse
+     - run: sudo mv wheelhouse/PyNaCl*.whl pynacl-wheelhouse/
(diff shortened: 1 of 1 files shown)


reaperhulk:
How long does this take in your tests?

odidev:
@reaperhulk, It is taking 41mins to complete aarch64 wheel builder pipeline.

odidev:
@reaperhulk , Could you please review the PR and give me your suggestions and feedback.
Thanks in advance.

reaperhulk:
This approach is too slow unfortunately. If we used this there would be a 40+ minute window where, after a release, a new sdist was available but no wheels would be. We do have access to graviton2 now that we've migrated our org, but we'll have to change our wheel building approach to happen automatically on tag (and sort out how we want to handle pypi tokens) to switch to that model.

ianw:
> This approach is too slow unfortunately. If we used this there would be a 40+ minute window where, after a release, a new sdist was available but no wheels would be. We do have access to graviton2 now that we've migrated our org, but we'll have to change our wheel building approach to happen automatically on tag (and sort out how we want to handle pypi tokens) to switch to that model.

I will be happy to work the existing OpenDev Zuul wheel builds into this repository as well, if that would be OK.  We can share most of the roles and make the jobs generic.  I think the Zuul bot would just need to be enabled here too.

Zuul could definitely upload sdists and linux wheels automatically on tags.  Unfortunately we don't have anywhere to build osx and windows though.  So I understand you want to keep everything together, but the offer is there :) 

> and sort out how we want to handle pypi tokens

FYI Zuul handles secrets for situations like this with the idea of post-review pipelines (https://zuul-ci.org/docs/zuul/reference/secret_def.html).  Zuul exports a key for each repo which you can encode a secret with (https://zuul-ci.org/docs/zuul/discussion/encryption.html); i.e. you keep the secret encrypted in your public job definitions and Zuul can decrypt it as it needs to.  You can only use that secret in trusted places -- i.e. the post-review release pipeline, that happens in response to tag events.  If a change has been tagged, then it's assumed only an authorized person has done that, and the tagged code has been reviewed and committed so is not doing anything silly like saying "hey, print out the value of this secret for me".  You obviously can not submit a pull request (that executes speculatively) that dumps a secret as that isn't a trusted context.

So in short, if the project added the openstackci pypi user, I could encrypt the secret I know for that, and we'd have a release pipeline job that would uploading using it.

odidev:
May I know, when the migration form travis-ci.org to travis-ci.com will be done? This PR is blocked and waiting for the migration to be completed.

reaperhulk:
It's already done. I think changing wheel building here looks something like this. To do a release we push a new tag. That tag triggers a workflow that does the following:
* Uploads the sdist as one job
* In parallel generates sdists from the checked out source from tag (we always want to work from a generated sdist) and then builds and uploads wheels

Even if the wheel building is slow this is okay because each wheel will be uploaded as they finish (instead of waiting until the slowest finishes in our current system).

This is probably easiest to implement with a minimum of changes on GitHub Actions, but would be faster on Travis since we have access to Graviton2 builders there.

odidev:
@reaperhulk I would be happy to add travis support for building wheels for pynacl. Would you like me to raise me a PR for the same?

odidev:
@reaperhulk, could you please let me know your interest on adding Travis-ci support for releasing wheels.

lopezm1:
@odidev Appreciate you following up and taking this on for the community. Our group just came across this issue and are interested in the release of it too.

odidev:
@lopezm1, thanks for your response. It seems, pynacl no longer uses travis-ci as CI/CD and I have already raised this PR for building/deploying aarch64 wheels. Can you please let me know what else is required here from my side. Thanks

odidev:
@lopezm1, can you please let me know if anything is required here from my side?
CC: @reaperhulk

reaperhulk:
We aren't going to merge any wheel builders that don't have corresponding CI, and we have no access to aarch64 hardware within GitHub Actions at this time. We could potentially look towards opendev (which we use on the cryptography side), but the reliability of that platform has been moderately frustrating for us.

odidev:
> aren't going to merge any wheel builders that don't have corresponding CI,

Please let me know your interest in running CI tests in qemu emulation environment. I will work on the same.

reaperhulk:
You implemented qemu based CI and in https://github.com/pyca/pynacl/pull/612#issuecomment-682595142 I noted that the performance is simply not acceptable for us. 40+ minute CI is not tenable.

joekohlsdorf:
I would agree that 40 minutes would be too much if you released multiple times per day but that is not the case at all.
Why would you make the sdist available on PyPI before all wheels are built? As far as I can see, uploading to PyPI is not part of the CI pipeline and nobody is pulling Python source releases from GitHub.

dennisameling:
FWIW, CircleCI has [free Linux arm64 runners](https://circleci.com/docs/2.0/arm-resources/) available since a few months 😊

reaperhulk:
@dennisameling thanks for the suggestion! I'm looking into that on `pyca/cryptography` based on your comment (https://github.com/pyca/cryptography/pull/6763). It appears to do what we need so once that lands we'll probably port it here as well.

reaperhulk:
This is now provided natively by circleci (until actions adds support for native arm at least)

PR #620, opened 2020-08-14
simplify test matrix

reaperhulk:
use focal for the builders
don't test different compilers (libsodium does that)
test system libsodium less (the permutations don't give better coverage)


alex:
pypy build is failing looks like.

alex:
travis failing

PR #624, opened 2020-08-27
new black, not quite the same as the old black

reaperhulk:


docs/vectors/python/argondriver.py
```diff
@@ -33,8 +33,7 @@ def __init__(self, args):
         self.count = 0
 
     def _runOnce(self, passwd, salt, dgst_len, maxmem, iters):
-        """
-        """
+        """"""
```
alex, line 36: Can we just delete this?

PR #629, opened 2020-09-13
Fix wheels

reaperhulk:


reaperhulk:
https://github.com/reaperhulk/pynacl/runs/1108658309?check_suite_focus=true (tested)

alex (approved):

Issue #639, opened 2020-11-06
Going from a string representation from a privatekey to a public key


BlackTurtle123:
I am trying to create a library for some crypto project.
They however require to be able to generate the public key, and the address from a privatekey string (which users would input), how is this done? I don't see any option for this currently is this right?

`privKey = SigningKey(accountSeedHash)
pubKey = privKey.verify_key
self.privateKey = base58.b58encode(privKey._signing_key)`

jakob:
If you only need the public key you can do it like this:

```python
import base58
import nacl.signing
import nacl.bindings

signing_key_b58 = "5rUDHu3Kf8LaLeW2Z2KJsLFMj3b8qzbJyn3GDxGJcxo8XDP1tj16aggSmdoQfVf45QB3uSby2J54Wbe3iKffVCxZ"
signing_key_bytes = base58.b58decode(signing_key_b58)

verify_key_bytes = nacl.bindings.crypto_sign_ed25519_sk_to_pk(signing_key_bytes)
verify_key = nacl.signing.VerifyKey(verify_key_bytes)
print(base58.b58encode(verify_key._key))
```

If you need the SigningKey object as well, you need to first extract the seed from the private key:
```python
seed = nacl.bindings.crypto_sign_ed25519_sk_to_seed(signing_key_bytes)
signing_key = nacl.signing.SigningKey(seed)
print(base58.b58encode(signing_key._signing_key))
 ```

You can read the relevant libsodium docs here:
https://doc.libsodium.org/public-key_cryptography/public-key_signatures

And the PyNaCl docs here:
https://pynacl.readthedocs.io/en/latest/signing/

Issue #419, opened 2018-03-27
It is awkward to instantiate a SigningKey from its value.


dmerejkowsky:
### Use case:

We have a signing keypair that has been generated by an other program in a different language, and we need to sign and verify messages with them.

The naive way does not work:

```python
message = b'this is a message'
raw_signing_key = b'....' # a bunch of bytes of length 64
signing_key = SigningKey(raw_signing_key, encoder=RawEncoder)
signing_key.sign(message)
```

That's because the SigningKey constructor expects a *seed*, not the value:

```python
# in src/nacl/signing.py
class SigningKey(...):

    def __init__(self, seed, encoder=encoding.RawEncoder):
        ...
        public_key, secret_key = nacl.bindings.crypto_sign_seed_keypair(seed)
```

We have a workaround, though:

We assing the `_signing_key` variable directly:

```python
message = b'this is a message'
raw_signing_key = b'....' # a bunch of bytes of length 64
fake_seed = b"\0" * 32
signing_key = nacl.signing.SigningKey(fake_seed)
signing_key._signing_key = raw_signing_key
signing_key.sign(message)
```

And then we can verify the signature by instantiating a VerifyKey directly:

```python
raw_verify_key = b'....' # bunch of bytes of length 32
verify_key = nacl.signing.VerifyKey(public_key, encoder=RawEncoder)
```

Of course, this is a bad idea because if you use `signing_key.verify_key` you get the wrong key ...

### Proposed fix

Maybe we could have something like this instead ?

```python
class SigningKey:
    @classmethod
    def from_value(signing_key):
           res = cls()
           res._signing_key = signing_key
           res.verify_key = VerifyKey(bindings.crypto_sign_ed25519_sk_to_pk(signing_key))

```
And then:

```python
signing_key = SigningKey.from_value(raw_signing_key)
```

Thoughts ?

lmctv:
@dmerejkowsky would you mind giving some more details about the raw keypairs you are dealing with, and in particular the environment in which they have been generated in the first place? Are the secret and the public part both available, or only the secret? Thank you.

dmerejkowsky:
The keys have been generated in Javascript (using libsodium-wrapper by the way).

The public and private keys are both available.

Since I opened the bug we thought more about the problem a bit more and discovered we could just use the lower-lever API in nacl.bindings directly:

```python
from nacl.bindings.crypto_sign import crypto_sign, crypot_sign_open, crypto_sign_BYTES

# generate a detached signature:
[...]

PR #648, opened 2020-12-11
Allow initializing SigningKey from a raw key

jakob:
Currently, `SigningKey` must be initialized from a 32 byte seed.

However, sometimes you already have the 64 byte Ed25519 private key, and not the seed. Then it's very inconvenient to create a `SigningKey` object from the key bytes.

This problem was initially reported by @dmerejkowsky in issue #419. Recently, @BlackTurtle123 opened issue #639, which is also about the same problem.

The problem is that most applications probably store the 64 byte Ed25519 key rather than the 32 byte seed, since that should be faster as you don't need to re-compute the public key every time.

This Pull Request adds an additional argument `key` to the `SigningKey` initializer. You can use this argument to create a `SigningKey` directly from the raw Ed25519 private key bytes.

The new signature proposed in this PR is:
`SigningKey(seed=None, encoder=encoding.RawEncoder, key=None)`

`key` is an optional 3rd argument in order to maintain backward compatibility with existing code.

Example usage of the new code:
```Python
raw_key = b'64 byte Ed25519 key from somewhere'
signing_key = SigningKey(key=raw_key)
```

### DIFF
--- a/src/nacl/signing.py
+++ b/src/nacl/signing.py
@@ -143,34 +143,61 @@ class SigningKey(encoding.Encodable, StringFixer, object):
     Signing keys are produced from a 32-byte (256-bit) random seed value. This
     value can be passed into the :class:`~nacl.signing.SigningKey` as a
     :func:`bytes` whose length is 32.
+    
+    Alternatively, you can construct :class:`~nacl.signing.SigningKey` directly
+    from a 64-byte Ed25519 private key. You must provide either a seed or a key,
+    but not both.
 
     .. warning:: This **must** be protected and remain secret. Anyone who knows
         the value of your :class:`~nacl.signing.SigningKey` or it's seed can
         masquerade as you.
 
-    :param seed: [:class:`bytes`] Random 32-byte value (i.e. private key)
+    :param seed: [:class:`bytes`] Random 32-byte value for generating Ed25519 private key
     :param encoder: A class that is able to decode the seed
+    :param key: [:class:`bytes`] A previously generated 64-byte Ed25519 private key
 
     :ivar: verify_key: [:class:`~nacl.signing.VerifyKey`] The verify
         (i.e. public) key that corresponds with this signing key.
     """
 
-    def __init__(self, seed, encoder=encoding.RawEncoder):
-        # Decode the seed
-        seed = encoder.decode(seed)
-        if not isinstance(seed, bytes):
-            raise exc.TypeError(
-                "SigningKey must be created from a 32 byte seed"
-            )
+    def __init__(self, seed=None, encoder=encoding.RawEncoder, key=None):
+        exc.ensure(
+            (key is not None and seed is None) or (key is None and seed is not None),
+            "SigningKey must be created either from a 32 byte seed or a 64 byte key."
+        )
+        if seed is not None:
+            # Decode the seed
+            seed = encoder.decode(seed)
+            if not isinstance(seed, bytes):
+                raise exc.TypeError(
+                    "SigningKey must be created from a 32 byte seed"
+                )
 
-        # Verify that our seed is the proper size
-        if len(seed) != nacl.bindings.crypto_sign_SEEDBYTES:
-            raise exc.ValueError(
-                "The seed must be exactly %d bytes long"
-                % nacl.bindings.crypto_sign_SEEDBYTES
-            )
+            # Verify that our seed is the proper size
+            if len(seed) != nacl.bindings.crypto_sign_SEEDBYTES:
+                raise exc.ValueError(
+                    "The seed must be exactly %d bytes long"
+                    % nacl.bindings.crypto_sign_SEEDBYTES
+                )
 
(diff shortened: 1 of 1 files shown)


dmerejkowsky:
Looks good to me - thanks

jakob:
An alternative would be to just check the length of the argument -- if it's 32 bytes, assume it's a seed, and if it's 64 bytes assume it's a full Ed25519 secret key (including the public key). That would also be backwards compatible, and may be a bit more intuitive. The version implemented in this PR requires reading the docs.

reaperhulk:
I think calling it key is misleading since the 32-byte value is called a key basically everywhere that isn't `libsodium`. `libsodium` itself uses secret key and seed, which is perhaps a reasonable path.

Other than that I'm okay with this approach.

jakob:
@reaperhulk OK. So should I change the argument from `key` to `secret_key`? We could also call it `sk`, like in the libsodium docs.

(Aside from that, I realised that I used the term "private key" in the docs. I guess I should probably replace that with "secret key".)

reaperhulk:
Let’s do secret_key

jakob:
OK, I've changed the argument name from `key` to `secret_key`.

I didn't change the term "Private Key" in the docs to "Secret Key" since "Private Key" is used everywhere in the docs.

reaperhulk:
Looks like this needs a black format

reaperhulk (commented):
Two final requests: instead of an assertion error I’d prefer a value error when the values are both None (you can do this with raising=exc.ValueError) and also please add a changelog entry. Thanks for working on this!

covert-encryption:
Should also accept 32 byte plain keys. This 64-byte combined key format of libsodium is quite annoying, and too often I end up doing hacks like `sk + bytes(32)` to keep pynacl happy when the function I am calling does not actually use the public key half at all.

jakob:
> Should also accept 32 byte plain keys. This 64-byte combined key format of libsodium is quite annoying, and too often I end up doing hacks like `sk + bytes(32)` to keep pynacl happy when the function I am calling does not actually use the public key half at all.

@covert-encryption Can you be more specific what you mean with "32 byte plain keys"? It sounds like you are talking about the value that PyNaCl and libsodium call "seed", so you could just create a `SigningKey(seed = [32 byte plain key])`

What this PR adds is allowing to create a SigningKey from the 64 byte value that libsodium calls a secret key (which is just the 32 byte seed and the 32 byte precomputed public key concatenated).

Tronic:
@jakob By plain key I mean the first 32 bytes of Ed25519 sk, i.e. the seed (which is SHA-512 hashed and clamped to obtain the actual value used in arithmetics). Some functions of libsodium require for the public key to be concatenated on it, thus you get 64 bytes. This is supposedly done for performance reasons, as it avoids seeding, clamping and scalarmult_base to obtain the public key. The problem is that many functions that don't actually use the public key at all still require 64 bytes, where they really should require only 32 bytes (which I believe is enough on the C API).

jakob:
OK, but does this have anything to do with this PR?

reaperhulk:
@jakob I can approve and merge this once the two small comments in the review are fixed!

jagerman:
The documentation in this PR is quite misleading:

> :param seed: [:class:`bytes`] Random 32-byte value for generating Ed25519 private key
> :param secret_key: [:class:`bytes`] A previously generated 64-byte Ed25519 private key

This seems to be implying that the seed "generates" something, but neither of these are used to "generate" anything, nor are they actually any different as the description is implying.  The starting point everywhere in libsodium is the seed and is really the only thing that ever should be stored.  The only difference is that, for computational convenience, libsodium wants to have a precomputed public key so that it doesn't have to be computed in many different functions.

I'm also pretty skeptical of the claim that

> it's very inconvenient to create a SigningKey object from the key bytes.

How does `seed=raw_key[:32]` qualify as "very inconvenient"?

jakob:
> How does seed=raw_key[:32] qualify as "very inconvenient"?

Neither libsodium nor pynacl mention in the docs that this is possible. You have to read the source code to know that. I didn't know that when I started on this PR.

It seems that libsodium tries to hide the fact that a secret key is just the concatenation of seed and publickey. Why else would they have functions like crypto_sign_ed25519_sk_to_seed?

jagerman:
> Neither libsodium nor pynacl mention in the docs that this is possible. You have to read the source code to know that. I didn't know that when I started on this PR.

That's fair.  I'd suggest at least a change to "secret key" instead of "private key" for the args/documentation then.  Private key immediately makes me think of the private key scalar, which *is* what a pynacl PrivateKey is (and, somewhat perversely, the PrivateKey that you get back from a `to_curve25519_private_key` actually *is also* the private key scalar for ed25519).

reaperhulk:
I apologize for how this didn't go anywhere. Since I'm not sure the status I'm going to go ahead and close it, but if you'd still like this then feel free to rebase and open a new PR and we can get it reviewed/merged.

PR #652, opened 2020-12-23
remove travis since it no longer works

reaperhulk:
We have no arm64 testing solution for the moment, but we'll get something set up again soon. We don't ship wheels yet so removing testing isn't a regression.


alex:
I'm very uncomfortable merging this as is -- it transitions us from "visibly broken CI" to "silently broken CI".

reaperhulk:
I don't follow. Travis tested only aarch64, an architecture for which we have never shipped a wheel on this project (nor do we have a wheel builder for it). This restores the status quo on architecture and we can separately re-add arm64 via opendev or some other means.

alex (approved):
Ok, that's my bad

PR #668, opened 2021-05-12
black fix and drop py2.7

reaperhulk:


alex:
py27 thingy is now a 404.... time to drop py2 here?

reaperhulk:
Yeah let's drop it.

pyproject.toml
```diff
@@ -9,4 +9,4 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 79
-target-version = ["py27"]
+target-version = ["py36"]
```
alex, line 12: This should be 3.5.

alex (changes requested):
Wheel builder update?

pyproject.toml
reaperhulk: I made it 36 on the assumption we'll drop 35 soon, but it's an easy change so sure 😄

alex (approved):

PR #676, opened 2021-06-27
Expose a high-level API for libsodium's AEAD functionality

nbraud:
- Add secret.Aead
- tests/secret:
  - Refactor, and randomize some of the tests
  - Add tests for secret.Aead

### DIFF
--- a/src/nacl/secret.py
+++ b/src/nacl/secret.py
@@ -180,7 +180,9 @@ class Aead(encoding.Encodable, StringFixer, object):
     KEY_SIZE = nacl.bindings.crypto_aead_xchacha20poly1305_ietf_KEYBYTES
     NONCE_SIZE = nacl.bindings.crypto_aead_xchacha20poly1305_ietf_NPUBBYTES
     MACBYTES = nacl.bindings.crypto_aead_xchacha20poly1305_ietf_ABYTES
-    MESSAGEBYTES_MAX = nacl.bindings.crypto_aead_xchacha20poly1305_ietf_MESSAGEBYTES_MAX
+    MESSAGEBYTES_MAX = (
+        nacl.bindings.crypto_aead_xchacha20poly1305_ietf_MESSAGEBYTES_MAX
+    )
 
     def __init__(self, key, encoder=encoding.RawEncoder):
         key = encoder.decode(key)
@@ -197,7 +199,9 @@ class Aead(encoding.Encodable, StringFixer, object):
     def __bytes__(self):
         return self._key
 
-    def encrypt(self, plaintext, aad=b'', nonce=None, encoder=encoding.RawEncoder):
+    def encrypt(
+        self, plaintext, aad=b"", nonce=None, encoder=encoding.RawEncoder
+    ):
         """
         Encrypts the plaintext message using the given `nonce` (or generates
         one randomly if omitted) and returns the ciphertext encoded with the
@@ -242,7 +246,9 @@ class Aead(encoding.Encodable, StringFixer, object):
             encoder.encode(nonce + ciphertext),
         )
 
-    def decrypt(self, ciphertext, aad=b'', nonce=None, encoder=encoding.RawEncoder):
+    def decrypt(
+        self, ciphertext, aad=b"", nonce=None, encoder=encoding.RawEncoder
+    ):
         """
         Decrypts the ciphertext using the `nonce` (explicitly, when passed as a
         parameter or implicitly, when omitted, as part of the ciphertext) and
--- a/src/nacl/secret.py
+++ b/src/nacl/secret.py
@@ -188,11 +188,18 @@ class Aead(encoding.Encodable, StringFixer, object):
         one randomly if omitted) and returns the ciphertext encoded with the
         encoder.
 
-        .. warning:: It is **VITALLY** important that the nonce is a nonce,
-            i.e. it is a number used only once for any given key. If you fail
-            to do this, you compromise the privacy of the messages encrypted.
-            Give your nonces a different prefix, or have one side use an odd
-            counter and one an even counter. Just make sure they are different.
+        .. warning:: It is vitally important for :param nonce: to be unique.
+            By default, it is generated randomly; [:class:`Aead`] uses XChacha20
+            for extended (192b) nonce size, so the risk of reusing random nonces
+            is negligible.  It is *strongly recommended* to keep this behaviour,
+            as nonce reuse will compromise the privacy of encrypted messages.
+            Should implicit nonces be inadequate for your application, the
+            second best option is using split counters; e.g. if sending messages
+            encrypted under a shared key between 2 users, each user can use the
+            number of messages it sent so far, prefixed or suffixed with a 1bit
+            user id.  Note that the counter must **never** be rolled back (due
+            to overflow, on-disk state being rolled back to an earlier backup,
+            ...)
 
         :param plaintext: [:class:`bytes`] The plaintext message to encrypt
         :param nonce: [:class:`bytes`] The nonce to use in the encryption
--- a/src/nacl/secret.py
+++ b/src/nacl/secret.py
@@ -33,6 +33,12 @@ class SecretBox(encoding.Encodable, StringFixer, object):
     send are repudiable. For non-repudiable messages, sign them after
     encryption.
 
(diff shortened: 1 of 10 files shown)


src/nacl/secret.py
```diff
@@ -132,3 +132,121 @@ def decrypt(self, ciphertext, nonce=None, encoder=encoding.RawEncoder):
...
+    def __bytes__(self):
+        return self._key
+
+    def encrypt(self, plaintext, aad=b'', nonce=None, encoder=encoding.RawEncoder):
+        """
+        Encrypts the plaintext message using the given `nonce` (or generates
+        one randomly if omitted) and returns the ciphertext encoded with the
+        encoder.
+
+        .. warning:: It is **VITALLY** important that the nonce is a nonce,
```
nbraud, line 191: I kept that from `SecretBox`, but TBH I don't particularly like that wording?
Encouraging users to manage nonces on their own, seems to me much more dangerous than risking randomly-generated nonces colliding (especially since XChacha20 uses 192b nonces, so users are safe with up to 2⁹⁶ messages encrypted under the same key and random nonces)

tests/test_secret.py
```diff
@@ -47,19 +48,37 @@
...
+    return st.binary(min_size=m.KEY_SIZE, max_size=m.KEY_SIZE).map(binascii.hexlify)
 
 
-def test_secret_box_bytes():
-    s = SecretBox(
-        b"ec2bee2d5be613ca82e377c96a0bf2220d823ce980cdff6279473edc52862798",
-        encoder=HexEncoder,
-    )
-    assert bytes(s) == s._key
+def box(m):
```
reaperhulk, line 55: Let's name this something else just to minimize confusion since there are also args below that are named box

reaperhulk (commented):
I'd like to more prominently document that this class is xchacha20poly1305 IETF under the hood. I'd also like to see round trip tests from the vectors. I think we already have those for the underlying bindings, but let's run them against this class as well.

reaperhulk:
And apologies for the very slow review :(

reaperhulk:
Looks like some flake8 failures here too

nbraud:
> I'd like to more prominently document that this class is xchacha20poly1305 IETF under the hood.

Done; while I was at it, I also documented the algorithm used in `SecretBox` and the (lack of) limits on the number and size of plaintexts (in both cases).

> I'd also like to see round trip tests from the vectors. I think we already have those for the underlying bindings, but let's run them against this class as well.

Yes, they exist at the level of the bindings, so it didn't seem terribly useful to repeat them, but I can do that if it's required.

nbraud:
@reaperhulk Done.

nbraud:
@reaperhulk Something I noticed making the changes you requested: I handled `encoder` the same way `SecretBox` does, but I think it would be a lot more ergonomic and intuitive if the `encoder` that was passed when constructing the `Aead` object was the default for `.encrypt` and `.decrypt`.

What do you think?

nbraud:
I cannot reproduce the CI failure under CPython 3.8 or 3.9  :/

nbraud:
Sorry, had to force-push to fix a typo in that last commit's title (and squashed the `fixup!` commit while I was at it)

nbraud:
Erf, of course I cannot use f-strings in 8d2e99b, as the testsuite is run against CPython 3.5  >_>'
FIxing that right away

nbraud:
Back to the same CI failure, which somehow only occurs under Py3.5  :/
I guess I can `print`-debug that by throwing temporary commits at the CI  >_>'

nbraud:
Interestingly, the debug `print` call causes `binascii.unhexlify` not to throw an exception, but now the test's parameters are shuffled around  >_>'

nbraud:
OK, fixed the issue: `dict` was reimplemented in Py3.6, and I accidentally relied on the new implementation being order-preserving

nbraud:
@reaperhulk Ping?

src/nacl/secret.py
```diff
@@ -33,6 +33,12 @@ class SecretBox(encoding.Encodable, StringFixer, object):
     send are repudiable. For non-repudiable messages, sign them after
     encryption.
 
+    Encryption is done using `XSalsa20-Poly1305`_, and there are no practical
+    limits on the number or size of messages (up to 2⁶⁴ messages, each up to 2⁶⁴
```
reaperhulk, line 37: I suspect using unicode like this will cause problems somewhere (sadly). Should probably just be 
```
2 \ :sup:`64`
```

src/nacl/secret.py
```diff
@@ -132,3 +138,143 @@ def decrypt(self, ciphertext, nonce=None, encoder=encoding.RawEncoder):
...
+    an exception.
+
+    Likewise, the authenticator should not be mistaken for a (public-key)
+    signature: recipients (with the ability to decrypt messages) are capable of
+    creating arbitrary valid message; in particular, this means AEAD messages
+    are repudiable. For non-repudiable messages, sign them after encryption.
+
+    The cryptosystem used is `XChacha20-Poly1305`_ as specified for
+    `standardization`_. There are `no practical limits`_ to how much can safely
+    be encrypted under a given key (up to 2⁶⁴ messages each containing up
```
reaperhulk, line 162: same

tests/test_secret.py
```diff
@@ -15,14 +15,19 @@
...
 import pytest
 
 from utils import flip_byte
 
 from nacl.encoding import HexEncoder
 from nacl.exceptions import CryptoError
-from nacl.secret import SecretBox
+from nacl.secret import Aead, SecretBox
+
+from test_aead import xchacha20poly1305_ietf_vectors
```
reaperhulk, line 30: `from .test_aead import xchacha20poly1305_ietf_vectors` should work without requiring changes to application-import-names

reaperhulk (commented):
almost there 😄

tests/test_secret.py
nbraud: I tried that first, and got `ImportError: attempted relative import with no known parent package`.

src/nacl/secret.py
nbraud: This change causes `tox -e meta` to fail with multiple `W605 invalid escape sequence '\ '` errors.

nbraud:
> @reaperhulk Something I noticed making the changes you requested: I handled `encoder` the same way `SecretBox` does, but I think it would be a lot more ergonomic and intuitive if the `encoder` that was passed when constructing the `Aead` object was the default for `.encrypt` and `.decrypt`.
> 
> What do you think?

@reaperhulk I think I missed that question.

tests/test_secret.py
reaperhulk: Oh we need to add an empty `__init__.py` in the tests dir so that it's treated as a package.

src/nacl/secret.py
reaperhulk: Sigh, okay then. I guess we'll see if unicode angers other things 😄

reaperhulk:
I'd rather entirely remove encoders! I think they're not a useful feature given how trivial it is to do it yourself as needed. However, I'd prefer consistency across the project as long as they're still around so even though it's less ergonomic let's keep it as is. (See #523 for a never-completed PR that deprecates encoders)

nbraud:
> I'd rather entirely remove encoders! I think they're not a useful feature given how trivial it is to do it yourself as needed. However, I'd prefer consistency across the project as long as they're still around so even though it's less ergonomic let's keep it as is.

Entirely valid, and I was expecting the outcome to either be “stay consistent” or “change other encoder-using things too”.  Thanks for the clarification that this is kinda-deprecated functionality.

tests/test_secret.py
nbraud: I didn't suggest that, as it involved changing all the test files (`from utils import ...` had to be changed to `.utils` and moved to the “first party” import group) but it's now done.

nbraud:
@reaperhulk I guess this should be good now.

nbraud:
@reaperhulk Given that you dropped Py3.5 support, I guess I should drop 623489ebd6e14744a0793f99cbe307d8eddf1269 and similar changes ?

reaperhulk:
Let’s just leave it as is for this PR. We can do an f string conversion in a future PR (although I don’t consider that a particularly high priority).

reaperhulk (approved):

PR #682, opened 2021-08-09
nacl.secret.Aead: Remove Py2 cruft (inheritence from object)

nbraud:


nbraud:
CI failure is spurious.

reaperhulk (approved):

PR #693, opened 2021-10-21
Add mypy to CI

DMRobertson:
Following the strategy outlined in https://github.com/pyca/pynacl/pull/692#issuecomment-948201912, I've

- Configured an initial, gentle version of mypy
- Added this to the CI in the meta step
- Fixed up mypy's initial complaints.


Dreamsorcerer:
> Not sure why the pypy CI build failed. Looks like it couldn't build `typed-ast` from source because `codecs.h` wasn't on the C include path?

Looks like there was a related issue on aiohttp, but I don't know the details: https://github.com/aio-libs/aiohttp/pull/5496

src/nacl/types.py
```diff
@@ -0,0 +1,13 @@
+import sys
+
+if sys.version_info >= (3, 8):
+    from typing import SupportsBytes
+else:
+    from typing_extensions import Protocol
+
+    class SupportsBytes(Protocol):
```
reaperhulk, line 8: What would be the disadvantage of just defining the types where you use `SupportsBytes` as just `bytes`? Clearly that makes it so types that can be coerced into bytes won't be detected, but is that a common case?

Dreamsorcerer: Looking at that again, I don't see any indication that `SupportsBytes` in 3.8+ only. It appears to be available in 3.6, so this should be an unconditional import.

Dreamsorcerer: Which would also mean that there's no need to have this file, just import from `typing`.

DMRobertson: > What would be the disadvantage of just defining the types where you use `SupportsBytes` as just `bytes`? Clearly that makes it so types that can be coerced into bytes won't be detected, but is that a common case?

No huge disadvantage; I just wanted to give mypy as complete a picture as possible, and avoid a `# type: ignore[foo]` unless necessary.

I only want to use this for `StringFixer` and `Encodable` (even in the big typedump of #692.) Not massively common.

Given that SupportsBytes is in 3.6 and up, I can just import directly as @Dreamsorcerer says and there's no need for any of this faff. How does that sound?

> Looking at that again, I don't see any indication that `SupportsBytes` in 3.8+ only. It appears to be available in 3.6, so this should be an unconditional import.

Thanks for spotting this. (I assumed that SupportsBytes wasn't available until 3.8 because that's when `Protocol` was made available; but it looks like [3.6 was using `_Protocol` internally](https://github.com/python/cpython/blob/2c56c97f015a7ea81719615ddcf3c745fba5b4f3/Lib/typing.py#L1777-L1782).

reaperhulk: I am happy with SupportsBytes since it supports 3.6+ without typing_extensions, yes

DMRobertson: Also, this was broken on 3.7 and below: `typing_extensions` wasn't marked as a requirement in `setup.py`.

reaperhulk (approved):

reaperhulk:
Thanks for working with me on this! I look forward to the next PR 😄 If at all possible it'd be great to split this into as many PRs as is reasonably feasible.

DMRobertson:
> Thanks for working with me on this! I look forward to the next PR smile If at all possible it'd be great to split this into as many PRs as is reasonably feasible.

My pleasure: thank you for having me. My plan is to tackle each module in a PR by itself, starting with #694.

PR #695, opened 2021-11-03
Type annotations for `nacl.encoding`

DMRobertson:
Pulled out from #692.

After we realised in #693 that we don't need `typing_extensions` to use
`SupportsBytes`, I didn't want to pull in `typing_extensions` again for
just one use of `Protocol`. (I have no other plans to use `Protocol` in
other annotations.)

My understanding: doing so means that we can't run `mypy` under Python
3.6 or 3.7 to fully typecheck the package. Maybe that's fine? I guess
the downside is that someone using `nacl` in a 3.6 or 3.7 project can't
fully benefit from these annotations. What're your thoughts here?

(As a compromise, I could do a `try-except` dance to use from
`typing_extensions.Protocol`, if it happens to be available.)

### DIFF
--- a/src/nacl/encoding.py
+++ b/src/nacl/encoding.py
@@ -98,7 +98,5 @@ class URLSafeBase64Encoder(_Encoder):
 
 
 class Encodable:
-    def encode(
-        self: SupportsBytes, encoder: Encoder = RawEncoder
-    ) -> bytes:
+    def encode(self: SupportsBytes, encoder: Encoder = RawEncoder) -> bytes:
         return encoder.encode(bytes(self))
--- a/src/nacl/encoding.py
+++ b/src/nacl/encoding.py
@@ -23,12 +23,15 @@ class _Encoder(metaclass=ABCMeta):
     @staticmethod
     @abstractmethod
     def encode(data: bytes) -> bytes:
-        pass
+        """Transform raw data to encoded data."""
 
     @staticmethod
     @abstractmethod
     def decode(data: bytes) -> bytes:
-        pass
+        """Transform encoded data back to raw data.
+
+        Decoding after encoding should be a no-op, i.e. `decode(encode(x)) == x`.
+        """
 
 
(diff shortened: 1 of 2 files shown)


reaperhulk:
I think doing a semi-gross abc workaround is probably the simplest answer here:

```python
class Encoder(metaclass=abc.ABCMeta)
@staticmethod
@abstractmethod
def encode(data: bytes) -> bytes:
    pass

@staticmethod
@abstractmethod
def decode(data: bytes) -> bytes:
    pass
```

And then make all the encoders subclasses of that. Uses subclassing (boo), but we can make things nicer and move to protocols when we get to drop 3.6/3.7 support.

DMRobertson:
@reaperhulk thanks for your input! I've done so in the latest commit (sorry for the force pushes there, I had some trouble with CI). I've prefixed the ABC's name with an underscore because I'm not sure if we want to expose it publicly. (Is the intention that users can define their own encoders?)

reaperhulk:
Looks like you need to run black. I’m happy with it being private. I'd like to entirely remove encoders as I don’t think they bring any real value given how simple it is to compose your own arbitrary encoding without the library, but that’s a future PR for someone to tackle.

src/nacl/encoding.py
```diff
@@ -11,73 +11,92 @@
...
+from typing import SupportsBytes, Type
+
+
+# TODO: when the minimum supported version of Python is 3.8, we can import
+# Protocol from typing, and replace Encoder with a Protocol instead.
+class _Encoder(metaclass=ABCMeta):
+    @staticmethod
+    @abstractmethod
+    def encode(data: bytes) -> bytes:
+        pass
```
reaperhulk, line 26: Coverage is upset about this, but we can entirely replace `pass` in these cases with a comment and it will be happy:

```python
@staticmethod
@abstractmethod
def encode(data: bytes) -> bytes:
    """
    The encode method
    """
```

Or some possibly less useless comment 😄

reaperhulk (approved):

PR #697, opened 2021-11-08
Type annotations for `nacl.utils`

DMRobertson:
This applies the changes proposed in #692 to `nacl.utils` on top of the
changes in more recent PRs.

I've annotated `nacl.bindings.randombytes_buf_deterministic` to appease
mypy. It's needed because I configured mypy to require that typed code
always calls typed functions, in #694.

Per https://github.com/pyca/pynacl/pull/692#discussion_r733872525 and
https://stackoverflow.com/a/44644576/5252017 , I've made sure that the
`_from_parts` accepts a generic `cls` argument and returns an instance
of that `cls`. I don't think we actually intend for people to subclass
`EncryptedMessage`, but maybe they do. Besides, it's nice to have
annotations that are as accurate as possible.

### DIFF
--- a/src/nacl/utils.py
+++ b/src/nacl/utils.py
@@ -19,7 +19,7 @@ from typing import SupportsBytes, Type, TypeVar
 import nacl.bindings
 from nacl import encoding
 
-C = TypeVar("C", bound="EncryptedMessage")
+_EncryptedMessage = TypeVar("_EncryptedMessage", bound="EncryptedMessage")
 
 
 class EncryptedMessage(bytes):
@@ -33,8 +33,8 @@ class EncryptedMessage(bytes):
 
     @classmethod
     def _from_parts(
-        cls: Type[C], nonce: bytes, ciphertext: bytes, combined: bytes
-    ) -> C:
+        cls: Type[_EncryptedMessage], nonce: bytes, ciphertext: bytes, combined: bytes
+    ) -> _EncryptedMessage:
         obj = cls(combined)
         obj._nonce = nonce
         obj._ciphertext = ciphertext
--- a/src/nacl/utils.py
+++ b/src/nacl/utils.py
@@ -14,10 +14,12 @@
 
 
 import os
-from typing import SupportsBytes
+from typing import SupportsBytes, Type, TypeVar
 
 import nacl.bindings
 from nacl import encoding
+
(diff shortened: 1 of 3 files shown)


DMRobertson:
`black`, `flake8` with `flake8-import-order` and `mypy` are all happy locally, so fingers crossed CI should be okay. :crossed_fingers:

src/nacl/utils.py
```diff
@@ -14,57 +14,63 @@
...
 
-    _nonce = object
-    _ciphertext = object
+    _nonce: bytes
+    _ciphertext: bytes
 
     @classmethod
-    def _from_parts(cls, nonce, ciphertext, combined):
+    def _from_parts(
+        cls: Type[C], nonce: bytes, ciphertext: bytes, combined: bytes
```
reaperhulk, line 36: Can we just pass cls: "EncryptedMessage" as a string here?

DMRobertson: I'm fine either way, but @Dreamsorcerer seemed pretty keen on this in #692. Perhaps they could advise?

Dreamsorcerer: Problem is that the code is not returning `EncryptedMessage()`, it's returning `cls()`. If you have a subclass, the return type is not `EncryptedMessage`, hence the need for a `TypeVar`.

I'd probably call it `_EncryptedMessage` or something though, to avoid any chance of someone accidentally reusing the type when it is bound to `EncryptedMessage`.

Dreamsorcerer (approved):

src/nacl/utils.py
reaperhulk: I’m not overly concerned about supporting subclassing of this type (don’t subclass things, this shouldn’t even be subclassing bytes, sigh), but I’m okay if we want to do it this way. Let’s do the private prefix if we keep it this way though.

Dreamsorcerer: > I’m not overly concerned about supporting subclassing of this type.

Sure, but my argument is that it should match the code. If we changed the return type to `EncryptedMessage`, then I expect to see the implementation use `EncryptedMessage`, rather than `cls`.

DMRobertson: The `Self` type proposed in [PEP 673](https://www.python.org/dev/peps/pep-0673/) would make this neater.

reaperhulk (approved):

PR #698, opened 2021-11-09
Tweak mypy settings

Dreamsorcerer:
Correct one option: https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_reexport

Add another warning option that might be useful in rare circumstances.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ target-version = ["py36"]
 [tool.mypy]
 show_error_codes = true
 check_untyped_defs = true
-implicit_reexport = false
+no_implicit_reexport = true
 warn_redundant_casts = true
 
 files = ["src/nacl"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ target-version = ["py36"]
 [tool.mypy]
 show_error_codes = true
 check_untyped_defs = true
-no_implicit_reexport = true
+implicit_reexport = false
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -14,7 +14,7 @@ target-version = ["py36"]
 [tool.mypy]
 show_error_codes = true
 check_untyped_defs = true
-no_implicit_reexport = true
+implicit_reexport = false
```
reaperhulk, line 17: https://mypy.readthedocs.io/en/stable/config_file.html#inverting-option-values

`no_implicit_reexport` is also a valid flag actually, but yours works as well.

Dreamsorcerer: Ah, so it is. Just found it weird that the documentation has an `implicit_reexport` and a `no_implicit_optional`.

reaperhulk: If you wouldn't mind reverting this I'm happy to merge the other change!

reaperhulk (approved):

DMRobertson:
I think this introduces the following `mypy` warnings:

```
pyproject.toml: [module = "nacl.bindings.crypto_aead"]: Per-module sections should only specify per-module flags (warn_incomplete_stub)
```

I think it just needs to be set under the first `[tool.mypy]` section.

Dreamsorcerer:
Ah, maybe I misunderstood the config format. Not sure why it still exits successfully with errors though.

PR #699, opened 2021-11-11
Type annotations for `nacl.bindings.crypto_aead`.

DMRobertson:
The next module in the `nacl` directory I looked at pulled in from `nacl.bindings`, so I thought that would be the next best place to work on.

Mypy config
-----------

Due to CFFI there are lots of `Any`s in play here. So first, adjust the
mypy config to tolerate `Any`s in the bindings module. The way I've done
this is a bit awkward. My first attempt included
`nacl.bindings.crypto_aead` in the existing block of overrides, together
with an additional override for `nacl.bindings.*` which relaxed the `Any`
checks. This doesn't work, and the [mypy docs](https://mypy.readthedocs.io/en/stable/config_file.html#config-file-format)
recognise this as being unfortunate.

I then tried to target the addtional override to just
`nacl.bindings.crypto_aead`, but this failed with:

```
pyproject.toml: toml config file contains [[tool.mypy.overrides]]
sections with conflicting values. Module 'nacl.bindings.crypto_aead' has
two different values for 'disallow_any_expr'
```

Annotations
-----------

Largely copied from #692, which itself cribbed types from docstrings.
There are two changes on top of that:

- I marked the constants at the top of the module as `int`s. They're
  `size_t`s in the corresponding C header. I'm not sure this is
  important (I think they get used with CFFI only?), but it drives down
  the number of `Any`s.
- The code suggested that `aead` was allowed to be an `Optional[bytes]`
  because we allow it to be `None` everywhere. I've changed the
  docstring and annotation to reflect this. I, erm, have no idea if
  passing `None`/`NULL` to the relevant functions is cryptographically
  meaningful or safe.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,14 +25,16 @@ module = [
 ]
 ignore_missing_imports = true
 
+# Most options can be applied across the entire project.
 [[tool.mypy.overrides]]
 module = [
+    "nacl.bindings.crypto_aead",
     "nacl.encoding",
     "nacl.exceptions",
     "nacl.utils",
 ]
 disallow_any_unimported = true
-disallow_any_expr = true
+# disallow_any_expr: true outside of `bindings`
 disallow_any_decorated = true
 disallow_any_explicit = true
 disallow_any_generics = true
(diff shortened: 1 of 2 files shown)


reaperhulk (approved):

reaperhulk:
We can probably do slightly larger PRs if that helps you out a bit. If they're < 400 lines of diff they're still reviewable 😄

PR #714, opened 2021-11-30
Annotations for `nacl.bindings.crypto_hash`

DMRobertson:
We're going to have fun resolving pyproject.toml later, I suspect.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ module = [
     "nacl.bindings.crypto_aead",
     "nacl.bindings.crypto_box",
     "nacl.bindings.crypto_core",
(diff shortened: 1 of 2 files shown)


reaperhulk:
You  might want to (temporarily) cheat by putting the new elements such that they aren't on adjacent lines. We can fix the ordering in a final PR, but that will prevent the conflicts. (This one has a conflict)

DMRobertson:
Thanks, that's a good idea.

> We can fix the ordering in a final PR, but that will prevent the conflicts. (This one has a conflict)

I'm hoping we can replace it with a `nacl.bindings.*` or `nacl.bindings` further down the line.

reaperhulk (approved):

PR #715, opened 2021-11-30
Annotations for `nacl.bindings.crypto_kx`

DMRobertson:
Again, going to need merge conflicts resolving later. I don't know if there's a better way to submit these changes?


reaperhulk:
See my previous comment about avoiding conflicts, but also I'm okay with the PRs being larger than this. We can do more modules when the changes are simple and the diff stays < 300 lines.

reaperhulk:
(conflict here too)

reaperhulk (approved):

PR #717, opened 2021-12-02
Mypy bindings part 2

DMRobertson:
Only one surprise here in c5fe5d32921c0f62cd5ca3ca0fb68d14bd950251. To quote from its commit message:

> Mypy couldn't see that `state.tagbuf is not None` in
> `crypto_secretstream_xchacha20poly1305_pull`, so I used a cast. I
> _think_ it's correct to treat this as an `Optional[bytes]`, because the
> docstrings mark many cffi `unsigned char[]` types as `bytes`. I'd
> appreciate a sanity check though!

In particular I'm not familiar with the mechanics of CFFI: My understanding is that the `buffer` objects it handles aren't regular `bytes` objects per-se, so I don't know if c5fe5d32921c0f62cd5ca3ca0fb68d14bd950251 is sound.

Alternatively, we could
 - type ignore the error mentioned in the quote
 - change the `ensure` call in that function to something that mypy can understand (i.e. a plain `if state.tagbuf is None: raise ...`)

### DIFF
--- a/src/nacl/bindings/crypto_secretstream.py
+++ b/src/nacl/bindings/crypto_secretstream.py
@@ -11,45 +11,45 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+from typing import Optional, Tuple, cast
 
 from nacl import exceptions as exc
 from nacl._sodium import ffi, lib
 from nacl.exceptions import ensure
 
 
-crypto_secretstream_xchacha20poly1305_ABYTES = (
+crypto_secretstream_xchacha20poly1305_ABYTES: int = (
     lib.crypto_secretstream_xchacha20poly1305_abytes()
 )
-crypto_secretstream_xchacha20poly1305_HEADERBYTES = (
+crypto_secretstream_xchacha20poly1305_HEADERBYTES: int = (
     lib.crypto_secretstream_xchacha20poly1305_headerbytes()
 )
-crypto_secretstream_xchacha20poly1305_KEYBYTES = (
+crypto_secretstream_xchacha20poly1305_KEYBYTES: int = (
     lib.crypto_secretstream_xchacha20poly1305_keybytes()
 )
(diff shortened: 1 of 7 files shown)


DMRobertson:
Codecov is unhappy, but I'm not sure how to interpret [its output](https://codecov.io/gh/pyca/pynacl/compare/5629937b4f7b127b33faa1af5c1c76e1ced05793...e890f7ed6ea66ed0445be0263bf0a765e4b34ef3)

reaperhulk:
This should be interpreted as codecov being bad. When I look at the commit it shows 100% as expected.

src/nacl/bindings/crypto_secretstream.py
```diff
@@ -317,10 +327,16 @@ def crypto_secretstream_xchacha20poly1305_pull(state, c, ad=None):
     )
     ensure(rc == 0, "Unexpected failure", raising=exc.RuntimeError)
 
-    return (ffi.buffer(state.rawbuf, mlen)[:], int(state.tagbuf[0]))
+    # Cast safety: we `ensure` above that `state.tagbuf is not None`.
+    return (
+        ffi.buffer(state.rawbuf, mlen)[:],
```
reaperhulk, line 332: Sorry, comment went on some random commit rather than this PR before. Is there an advantage to a cast here rather than just asserting not none so myth understands the type narrowing?

DMRobertson: No strong opinions here! I don't think there's advantage as such; I was just reluctant to disrupt the pattern of using `ensures` across the library. This seemed like a less disruptive change.

More generally, I suppose casts can
- narrow when working with types that only exist at typecheck time, (e.g. `Protocol`s that aren't marked as `runtime_checkable`
-- avoid runtime checks that may occur if asserts are enabled.
 
E.g. I could `cast(Dict[str, int], my_dict)` instead of doing
```python
for k, v in my_dict.items():
    assert isinstance(k, str)
    assert isinstance(v, int)
```

(I'm not sure if mypy is able to narrow based on that loop though.)

reaperhulk (approved):

PR #718, opened 2021-12-07
Annotations for `nacl.pwhash`.

DMRobertson:
Two commits: one which fixes incorrect annotations in `nacl.bindings`, and then the changes from #692 which apply to `nacl.pwhash`. It looks like these were mostly a case of transcribing the types in docstrings to the annotation syntax.

### DIFF
--- a/src/nacl/pwhash/scrypt.py
+++ b/src/nacl/pwhash/scrypt.py
@@ -56,13 +56,13 @@ MEMLIMIT_MODERATE = 8 * MEMLIMIT_INTERACTIVE
 
 
 def kdf(
-    size,
-    password,
-    salt,
-    opslimit=OPSLIMIT_SENSITIVE,
-    memlimit=MEMLIMIT_SENSITIVE,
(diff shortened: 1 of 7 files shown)


reaperhulk (approved):

src/nacl/pwhash/scrypt.py
```diff
@@ -138,8 +138,10 @@ def kdf(
 
 
 def str(
-    password, opslimit=OPSLIMIT_INTERACTIVE, memlimit=MEMLIMIT_INTERACTIVE
-):
+    password: bytes,
+    opslimit: int = OPSLIMIT_INTERACTIVE,
+    memlimit: int = MEMLIMIT_INTERACTIVE,
+) -> bytes:
```
reaperhulk, line 144: `str` which returns `bytes` 😆

reaperhulk:
The reason `ffi.string` is called that is because it assumes it's a C string (and is therefore null terminated). Confusing for `str` vs `bytes` in Python though, I agree.

PR #725, opened 2021-12-22
Annotations in `tests`, part 1

DMRobertson:
~~Draft for now while CI runs. ~~

I started by adding annotations to `tests.utils` like I did the rest of the project, using the same strict config as before. Then I took a few modules and ran mypy on them, and dealt with any unexpected complaints reported by mypy. Then I suppressed the uninteresting errors in mypy config.

Its complaints so far:
* test functions aren't annotated. For simple functions this is just `disallow_untyped_defs` being unhappy that we haven't marked test function as `-> None`. There are some parameterised tests though, and I found it useful to add in annotations here to ensure the test body was doing something sensible.
  * FWIW I wouldn't be against adding the `-> None`
* decorators from pytest and hypothesis return functions whose type involves `Any`. Not interesting; ignored in config.
* Mypy spotted that `crypto_scalarmult_BYTES` and `SCALARBYTES` were untyped, see 10434eb. Easy to fix.
* Also in 10434eb, we were calling `crypto_box` with a string `nonce`. I'm a bit surprised this worked, to be honest. I guess the fact that the nonce was ascii-encodable (i.e. just nulls) might have helped? 

I didn't feel like annotating `check_type_error` three times, so I pulled it out to `tests.utils`. It's a bit naughty to do so in this PR (could spin it out?) and I felt a bit icky making `utils` pull in `pytest`.

### DIFF
--- a/tests/test_box.py
+++ b/tests/test_box.py
@@ -23,7 +23,7 @@ from nacl.public import Box, PrivateKey, PublicKey
 from nacl.utils import random
 
 from .test_bindings import _box_from_seed_vectors
-
+from .utils import check_type_error
 
 VECTORS = [
     # privalice, pubalice, privbob, pubbob, nonce, plaintext, ciphertext
@@ -340,12 +340,6 @@ def test_box_wrong_length():
         b.encrypt(b"", b"")
     with pytest.raises(ValueError):
         b.decrypt(b"", b"")
-
-
-def check_type_error(expected, f, *args):
-    with pytest.raises(TypeError) as e:
-        f(*args)
-    assert expected in str(e.value)
 
 
 def test_wrong_types():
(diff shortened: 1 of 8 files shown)


tests/test_box.py
```diff
@@ -285,6 +323,7 @@ def test_box_failed_decryption(
 def test_box_wrong_length():
     with pytest.raises(ValueError):
         PublicKey(b"")
+    # TODO: should the below raise a ValueError?
```
reaperhulk, line 326: A good one to look at in the future, but fine to preserve behavior for now.

reaperhulk (approved):
I'm okay with utils having pytest since it's in the tests package 😄

reaperhulk:
Huge thanks for working through this big project with me. I think this is the last thing I would consider a blocker for the next pynacl release, so now it's just some housekeeping and getting a release out the door in the next few days/weeks.

DMRobertson:
> Huge thanks for working through this big project with me.

My pleasure---thanks for reviewing!

H-Shay:
Looks like pynacl 1.5.0 is out: https://github.com/pyca/pynacl/releases/tag/1.5.0, so we can probably go ahead with this now.

PR #730, opened 2021-12-23
Mypy tests 3

DMRobertson:
Also a draft. Branches off from #728; the first new commit is 
8cfeced. I'm just sticking this up to get CI to run on the proposed final batch of annotations to `tests` now.

Also, while I'm at it, I wanted to see if we could quantify how much has changed with all this typing. To do so:

```
git checkout mypy-tests-3
mypy --html-report out --any-exprs-report out
xdg-open out/index.html
less out/any-exprs.txt

git checkout ee93f9f13e4ce803a707b972d0fbc324439e71fc
git restore --source=mypy-tests-3 pyproject.toml
mypy --html-report out_old --any-exprs-report out_old
xdg-open out_old/index.html
less out_old/any-exprs.txt
```

TL:DR:

* Before: 55% of expressions were not Any. Mypy deemed the project as a whole 33% imprecise.
* After: 83% of expressions are not Any. Project as a whole deemed 12% imprecise.

(These numbers from the any-exprs and html reports, respectively).

I think most of that comes down to cffi in `nacl`; maybe some bits in tests from `json.loads()`.

If we just look at the higher level stuff outside of `nacl.bindings`, the numbers are rosier:

```
$ less out/any-exprs.txt | \grep nacl | \grep -v bindings 
                             nacl      0      19    100.00%
                    nacl.encoding      0      70    100.00%
                  nacl.exceptions      0      27    100.00%
                        nacl.hash      0     105    100.00%
                     nacl.hashlib      0     137    100.00%
                      nacl.public      5     489     98.98%
                      nacl.pwhash      0      85    100.00%
              nacl.pwhash._argon2      0      31    100.00%
              nacl.pwhash.argon2i      0      71    100.00%
             nacl.pwhash.argon2id      0      71    100.00%
               nacl.pwhash.scrypt      1     129     99.22%
                      nacl.secret      2     276     99.28%
                     nacl.signing      6     278     97.84%
                       nacl.utils      0      59    100.00%
```

### DIFF
--- a/src/nacl/bindings/crypto_secretstream.py
+++ b/src/nacl/bindings/crypto_secretstream.py
@@ -73,13 +73,15 @@ class crypto_secretstream_xchacha20poly1305_state:
 
     def __init__(self) -> None:
         """Initialize a clean state object."""
-        self.statebuf = ffi.new(
+        # NOTE: the members below aren't `bytearray` objects, but cffi `cdata` objects
+        # which own an array of unsigned chars. Is there a better annotation?
+        self.statebuf: bytearray = ffi.new(
             "unsigned char[]",
             crypto_secretstream_xchacha20poly1305_STATEBYTES,
         )
 
-        self.rawbuf: Optional[bytes] = None
-        self.tagbuf: Optional[bytes] = None
+        self.rawbuf: Optional[bytearray] = None
+        self.tagbuf: Optional[bytearray] = None
 
 
 def crypto_secretstream_xchacha20poly1305_init_push(
--- a/src/nacl/bindings/crypto_secretstream.py
+++ b/src/nacl/bindings/crypto_secretstream.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Optional, Tuple, cast
+from typing import ByteString, Optional, Tuple, cast
 
 from nacl import exceptions as exc
 from nacl._sodium import ffi, lib
@@ -73,15 +73,13 @@ class crypto_secretstream_xchacha20poly1305_state:
 
     def __init__(self) -> None:
         """Initialize a clean state object."""
-        # NOTE: the members below aren't `bytearray` objects, but cffi `cdata` objects
-        # which own an array of unsigned chars. Is there a better annotation?
-        self.statebuf: bytearray = ffi.new(
+        self.statebuf: ByteString = ffi.new(
             "unsigned char[]",
             crypto_secretstream_xchacha20poly1305_STATEBYTES,
         )
 
-        self.rawbuf: Optional[bytearray] = None
-        self.tagbuf: Optional[bytearray] = None
+        self.rawbuf: Optional[ByteString] = None
+        self.tagbuf: Optional[ByteString] = None
 
 
 def crypto_secretstream_xchacha20poly1305_init_push(
--- a/tests/test_sealed_box.py
+++ b/tests/test_sealed_box.py
@@ -14,6 +14,7 @@
 
 
 import binascii
+from typing import List, Tuple
 
 import pytest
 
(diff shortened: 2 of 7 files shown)


DMRobertson:
Codecov complains, but AFAICS there's no complaint here: https://app.codecov.io/gh/pyca/pynacl/compare/730?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr%20comments&utm_term=pyca

src/nacl/bindings/crypto_secretstream.py
```diff
@@ -73,13 +73,15 @@ class crypto_secretstream_xchacha20poly1305_state:
 
     def __init__(self) -> None:
         """Initialize a clean state object."""
-        self.statebuf = ffi.new(
+        # NOTE: the members below aren't `bytearray` objects, but cffi `cdata` objects
+        # which own an array of unsigned chars. Is there a better annotation?
```
DMRobertson, line 77: I considered `typing.ByteString` here, but 
- that's immutable, but one of the test writes to it
-  the docs [suggest](https://docs.python.org/3/library/typing.html?highlight=typing#typing.ByteString) that `bytes` in an annotation is really shorthand for a ByteString anyway.
    >  class typing.ByteString(Sequence[int])
    > 
    >   A generic version of collections.abc.ByteString.
    >
    >    This type represents the types bytes, bytearray, and memoryview of byte sequences.
    >
    >    As a shorthand for this type, bytes can be used to annotate arguments of any of the types mentioned above.

`bytearray` seemed like the least bad option, but would welcome input here!

tests/test_sealed_box.py
```diff
@@ -111,20 +115,23 @@ def test_wrong_types():
...
 ):
-    pubalice = PublicKey(pubalice, encoder=HexEncoder)
+    pubalice_decoded = PublicKey(pubalice, encoder=HexEncoder)
 
-    box = SealedBox(pubalice)
+    box = SealedBox(pubalice_decoded)
     with pytest.raises(TypeError):
-        box.decrypt(
+        # Type safety: mypy spots that you can't decrypt with a public key, but we
+        # want to detect this at runtime too.
```
DMRobertson, line 133: I was pleased to see this caught by mypy---I'd forgotten that we did this in #706!

src/nacl/bindings/crypto_secretstream.py
reaperhulk: In cryptography we've chosen to leave these untyped for now with a hope of typing `ffi.new` itself in the future. I suppose if we do that then we'd catch these cases here, but I'm not hugely comfortable with picking a "closest approximation" type. e.g. you can't call `bytes()` on this buffer, which you could on a `bytearray`.

DMRobertson: That makes sense.

The only place (AFAICS) that the statebuf gets mutated by Python is [here](https://github.com/DMRobertson/pynacl/blob/6c4f3a2bcb18abc20bf294ed12d1b8f56bfdcf6f/tests/test_secretstream.py#L242-L245), and that's just to copy the data from another ffi `unsigned char[]`.

Perhaps it would be safer to use `Sequence[int]` here instead of `bytearray` and type-ignore the part of the test that writes to this buffer. (I stuck a breakpoint in at the snippet linked above, where `state_save[0]` was the int `68` as opposed to `b"\x68"`.)

DMRobertson: Actually, `ByteString` seems like a better fit: it inherits from `Sequence[int]`, but additionally expresses that the ints are all going to be byte-sized.

DMRobertson: I've had a go at this in bc42daf.

reaperhulk (approved):

reaperhulk:
I'm going to avoid further nitpicking here since we can always find ways to improve in the future. 😁

DMRobertson:
For what it's worth, I'm open to nitpicking. Either way, great to see this merged and rounded off!

PR #766, opened 2022-06-09
Fix documentation of crypto_sign_ed25519ph_final_create

KloolK:
The description said that `sk` is the secret part only, where actually the full secret (secret part + public part) is expected.

I'm a little unsure about the correct terminology however. Feel free to adapt :)


reaperhulk (approved):

reaperhulk:
Thanks!

PR #792, opened 2023-05-02
Feature/expose from uniform

AnonimaUzanto:
Exposed nacl.bindings.crypto_core_ed25519_from_uniform from the low-level ed25519 API. The function accepts 32 random bytes and produces a valid point in the ed25519 main subgroup.

[Libsodium: crypto_core_ed25519_from_uniform documentation](https://doc.libsodium.org/advanced/point-arithmetic#elligator-2-map)

[Libsodium: crypto_core_ed25519_from_uniform testing logic](https://github.com/jedisct1/libsodium/blob/master/test/default/core_ed25519.c#L78)


reaperhulk:
#793 has fixed the sphinx issue so you can rebase this now.

AnonimaUzanto:
Rebased code and fixed flake issues.

tests/test_bindings.py
```diff
@@ -492,6 +493,23 @@ def test_ed25519_is_valid_point():
     assert res is False
 
 
+@pytest.mark.skipif(
+    not c.has_crypto_core_ed25519, reason="Requires full build of libsodium"
+)
+def test_ed25519_from_uniform():
```
reaperhulk, line 499: Are there any test vectors we can add here?

AnonimaUzanto: After significant searching I have not found canonical test vectors for ed25519. There are test [vectors](https://elligator.org/vectors/) for curve 25519 rather than ed25519, but those vectors are not compatible with the libsodium implementation. The IRTF document [examples](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-16#name-edwards25519-2) are also not compatible with the libsodium implementation.

The closest I can find to a comparison, though not validated, is present in a [comment](https://github.com/jedisct1/libsodium/discussions/1086) for libsodium. The current implementation matches the comment input/output:

in : 7f3e7fb9428103ad7f52db32f9df32505d7b427d894c5093f7a0f0374a30641d
out: 44b2fa2a6bb0b2adeace690a5a83b7fbe5bb487c34e64dc109b90bc4e00f670b

reaperhulk: Thanks for checking -- let's add one more test case that runs that libsodium test (with a comment noting its provenance) and then I guess we'll be set.

AnonimaUzanto: I modified the test with the following:
- added a check to confirm example input **is not** a valid curve point
- added a check to confirm from_uniform output **is** a valid curve point
- added a check to confirm generated output equals expected output

Also added a comment referencing the libsodium discussion for the example data.

reaperhulk (approved):

PR #820, opened 2024-05-13
Try to upgrade ubuntu version in CI

alex:


alex:
Sigh. I think it's time to drop Python 3.6, but I'm not going to do it myself today.

PR #830, opened 2024-07-17
Updated docs including Aead

gonatienza:
Aead from nacl.secret added to the secret docs.


reaperhulk (approved):
Thanks for doing this work -- it looks like when this feature landed it was only documented via docstring 😞

PR #837, opened 2024-09-01
restrict setuptools to work around breakage

reaperhulk:


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [build-system]
 # Must be kept in sync with `dependencies` below
 requires = [
-    "setuptools>=61.0.0",
+    "setuptools!=74.0.0",
```
alex, line 4: Probably we still want the minimum

alex (approved):

PR #845, opened 2024-10-20
Add license to package classiers

Rogdham:
Adding the license in the list of [package classifiers](https://pypi.org/classifiers/).

This is helpful for automated tools checking the license of packages, because the classifiers are a well-defined list, and tend to be well respected. This is contrary to the `license` metadata (which `PyNaCl` populates with `Apache-2.0` already), where the values used in various packages seems be all over the place.


alex (approved):

Rogdham:
Thanks for the merge! :hugs:

Issue #707, opened 2021-11-21
Should support any buffer objects, not just bytes


covert-encryption:
Use `ffi.from_buffer(...)` for conversions, so that `bytearray`, `memoryview` and other things work too, instead of checking that the arguments are `bytes` instances.

I would also prefer being able to give output buffers to low level functions, for instance I had to do this hack to enable encryption without creating a new buffer. It works even in place, if `ciphertext` and `message` are the same buffer:

```python
from nacl._sodium import ffi, lib

def encrypt_into(ciphertext, message, aad, nonce, key):
  mlen = len(message)
  clen = ffi.new("unsigned long long *")
  ciphertext = ffi.from_buffer(ciphertext)
  message = ffi.from_buffer(message)
  if aad:
    _aad = ffi.from_buffer(aad)
    aalen = len(aad)
  else:
    _aad = ffi.NULL
    aalen = 0

  return lib.crypto_aead_chacha20poly1305_ietf_encrypt(
    ciphertext, clen, message, mlen, _aad, aalen, ffi.NULL, nonce, key
  )
```

While for most things it does not matter if copies are made and new buffers are being allocated, stream ciphers can be much faster with proper buffer management.

reaperhulk:
Yes, this library predates `from_buffer` in `cffi`, but it should use it. `pyca/cryptography` switched long ago but no one has done the work on `pynacl` yet.

PR #847, opened 2024-11-12
Ensure that secretbox can accept any bytes-like object as parameter

blast-hardcheese:
start to address #707, ensuring that secretbox can accept `bytearray`.

The issue I ran into in this case was that both mypy and pyright think that `bytearray` satisfy `bytes`. Taking the suggestion from #707 resolved this in tests.

Thank you for the library!

### DIFF
--- a/src/nacl/bindings/crypto_secretbox.py
+++ b/src/nacl/bindings/crypto_secretbox.py
@@ -44,6 +44,9 @@ def crypto_secretbox(message: bytes, nonce: bytes, key: bytes) -> bytes:
     if len(nonce) != crypto_secretbox_NONCEBYTES:
         raise exc.ValueError("Invalid nonce")
 
+    nonce = ffi.from_buffer(nonce)
+    key = ffi.from_buffer(key)
+
     padded = b"\x00" * crypto_secretbox_ZEROBYTES + message
     ciphertext = ffi.new("unsigned char[]", len(padded))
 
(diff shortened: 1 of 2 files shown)


ryantm (approved):

blast-hardcheese:
Please pardon the force pushes. I'm also happy to remove mypy from `optional-dependencies.tests` if desired.

I was able to use [uv](https://docs.astral.sh/uv/) to do most of my iteration here, but not having an explicit version of `mypy` specified prevented typechecking locally.

reaperhulk:
@blast-hardcheese are you interested in implementing this across the set of methods? If so, we should set up an issue to track everything and burn it down. I've been meaning to ship a new PyNaCl release for a while, so it'd be nice to get this in there 🙂

blast-hardcheese:
I'm not opposed to it, I figured I'd start with the ones that were causing me pain. Is it sufficient to apply this pattern to all public methods?

reaperhulk:
Yes, this pattern is correct for any immutable buffers being passed into the C layer.

reaperhulk:
A very long time later I'm back around looking at these things. I'd still like to fix this systemically and I'll be happy to ship a release quickly (really!) if there's still interest in doing this for all immutable buffers. We're doing 1.6.0 shortly, but 1.7.0 with this is not a problem.

blast-hardcheese:
I've actually moved on, and I no longer have context on this. Python typechecking is an interesting space, so no doubt just running both mypy and pyright and seeing what the current state of the world looks like is a suitable next step, but I really don't know.

Thank you for the library, it was helpful while I was on that project!

PR #848, opened 2024-11-13
Fix compatibility with Python 3.14.0a1

befeleme:
typing.ByteString has been removed from Python 3.14. Ported the suggested way from the documentation:
https://docs.python.org/3.13/library/typing.html#typing.ByteString

I've verified the fix presented makes the package build and tests pass in Fedora Linux build environment: https://copr.fedorainfracloud.org/coprs/g/python/python3.14/package/python-pynacl/


alex (approved):

reaperhulk:
As soon as the cffi ft PR merges I intend to do a release.

PR #863, opened 2025-04-04
try to fix wheel-builder workflow

ngoldbaum:
It looks like the wheel filename is getting normalized now?

```
removing build/bdist.linux-x86_64/wheel
Successfully built pynacl-1.6.0.dev1-cp37-abi3-linux_x86_64.whl
mv: can't rename 'dist/PyNaCl*.whl': No such file or directory
```


alex:
Looks like this applies to the windows wheel builder as well

alex (approved):

alex:
thanks for jumping on this. (will merge after I get back from lunch)

PR #865, opened 2025-04-17
CI: Stop using ubuntu-20.04

lazka:
It's going away: https://github.com/actions/runner-images/issues/11101

Switch to ubuntu-22.04 instead. Switching to ubuntu-24.04 is only blocked
by Python 3.7 not being available there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   linux:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:


reaperhulk:
Any reason we shouldn't switch to `ubuntu-24.04`?

lazka:
> Any reason we shouldn't switch to `ubuntu-24.04`?

No, just tried to keep the changes minimal. Let me try

lazka:
Python 3.7 is not supported with 24.04: https://github.com/lazka/pynacl/actions/runs/14524276612/job/40752241633 Everything else worked.

```
Run actions/setup-python@v5
Installed versions
  Version 3.7 was not found in the local cache
  Error: The version '3.7' with architecture 'x64' was not found for Ubuntu 24.04.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

reaperhulk:
Ah yes, on pyca/cryptography we solved this by having a variety of containers we use that have 3.7 in them still. Okay, 22.04 is fine for now here, although we should consider dropping 3.7 support here soon-ish.

reaperhulk (approved):

lazka:
thanks!

PR #866, opened 2025-04-21
bump to libsodium-1.0.20-stable (03-Apr-2025)

reaperhulk:
This also adds arm64 windows builds to the msvc zip


alex:
Looks like we need to have ruff ignore the libsodium dir (caan you do that in a seperate PR plz)

alex (approved):

PR #870, opened 2025-05-07
typo

gonatienza:
Hey @reaperhulk ,

Just noticed this minor cosmetic typo.  Might affect any future docstring parsing by sphinx.

Glad all of those bindings were merged.  Thanks for taking the time bud!

### DIFF
--- a/tests/test_aead.py
+++ b/tests/test_aead.py
@@ -21,6 +21,7 @@ import pytest
 
 import nacl.bindings as b
 import nacl.exceptions as exc
+from nacl._sodium import lib
 
 from .utils import read_kv_test_vectors
 
@@ -95,6 +96,8 @@ def _getconstruction(construction: bytes) -> Construction:
         NPUB = b.crypto_aead_aegis128l_NPUBBYTES
         KEYBYTES = b.crypto_aead_aegis128l_KEYBYTES
     elif construction == b"aes256gcm":
+        if lib.crypto_aead_aes256gcm_is_available() != 1:
+            pytest.skip("aes256gcm not supported")
         encrypt = b.crypto_aead_aes256gcm_encrypt
         decrypt = b.crypto_aead_aes256gcm_decrypt
         NPUB = b.crypto_aead_aes256gcm_NPUBBYTES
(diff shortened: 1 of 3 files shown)


gonatienza:
Addressed other CI errors and squashed all the changes.

HTH.

tests/test_aead.py
```diff
@@ -123,7 +123,7 @@ def test_variants_kat(kv: Dict[str, bytes]):
     try:
         out = c.encrypt(msg, ad, nonce, k)
         assert out == exp
-    except exc.UnavailableError:
+    except exc.UnavailableError:  # pragma: no cover
```
reaperhulk, line 126: Didn't we have coverage here before? CI passed, so this is surprising. 😄 Generally I'd prefer to restructure the code to avoid branches we can't cover (e.g., we could use a pytest skipif decorator here)

gonatienza: Ya, there were likely some changes after that PR cleared.  Will make this PR cleaner later tonight.

gonatienza: The coverage error manifested itself without consistency.  Regardless, I made it better to ensure I don't catch the exceptions if the specific construct is unsupported.  Feels better to me, let me know if it works for you.

gonatienza:
Hey @reaperhulk , let me know if that works better.  Thanks!

tests/test_aead.py
gonatienza: The inconsistency of the coverage gap kept me thinking.  Maybe some of the resources running the actual CI tests didn't have AES acceleration available and actually hit the exception.  Hence it was actually fully covered in those cases.  That would explain why it presented itself randomly.

reaperhulk: That would be extraordinarily surprising, as AESNI support has been effectively universal in x86_64 since ~2010-2011 and Azure isn't going to be running any CPUs that old (nor any AMD CPUs of similar vintage). Regardless, the new approach will let us know if we ever encounter the issue and makes much more sense to me, thanks!

reaperhulk: Oh no it's much worse...this repo has never been updated to use the coverage combining system we use elsewhere and has a broken codecov deployment still. Guess I'll fix that next 😄

reaperhulk (approved):

PR #881, opened 2025-08-04
Use cffi-ft

clin1234:


reaperhulk:
The authors of that fork have gotten their work mered into cffi itself and the PR testing against that is here: https://github.com/pyca/pynacl/pull/880

PR #890, opened 2025-09-10
upgrade to libsodium 1.0.20-stable 2025-08-27 build

reaperhulk:


reaperhulk:
Libsodium has a 1.0.20 and a 1.0.20-stable. The latter is a rolling release without versioning but incorporates a variety of fixes over time. We have a 1.0.20-stable from April, this is the latest. It’s annoying.

PR #948, opened 2026-05-10
fix mypy bytearray-vs-bytes error in crypto_aead

greateggsgreg:
Fix mypy build errors that do not alter the spirit of the tests affected

### DIFF
--- a/tests/test_aead.py
+++ b/tests/test_aead.py
@@ -159,7 +159,7 @@ def test_variants_roundtrip_aad(
     with pytest.raises(exc.CryptoError):
         ct1 = bytearray(ct)
         ct1[0] = ct1[0] ^ 0xFF
-        c.decrypt(ct1, aad, unonce, ukey)
+        c.decrypt(bytes(ct1), aad, unonce, ukey)
 
 
 @given(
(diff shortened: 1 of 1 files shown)


alex:
I don't think this is the right fix -- this issue is caused by a behavioral chance in mypy 2.0. Previously `bytes` was actually an alias for `bytes | bytearray | memoryview`. In 2.0 it's strictly enforced. I think the thing we actually want to do is: a) verify if these functions should accept teh wider variety of types, and if so update the signatures.

greateggsgreg:
> I don't think this is the right fix -- this issue is caused by a behavioral chance in mypy 2.0. Previously `bytes` was actually an alias for `bytes | bytearray | memoryview`. In 2.0 it's strictly enforced. I think the thing we actually want to do is: a) verify if these functions should accept teh wider variety of types, and if so update the signatures.

Completely fair. After testing, this actually seems like the best fix.

I tried widening the type hint and testing bytesarray + bytes + memoryview - only bytes is valid unless we coerce inputs to bytes() for cffi. We could go the direction and widen to accepting bytes-like as a convenience, but this would be a big diff across the entire codebase.

Attached is a diff for what it would look like if we did go down the type widening route - note that we'd have to cast the inputs to bytes() anyway: [pynacl-wider-typing.diff.txt](https://github.com/user-attachments/files/27571332/pynacl-wider-typing.diff.txt)

alex:
this got resolved in a different PR, thanks for contributing.

